PR #125, opened 2019-11-20
fixed segment extraction

lorenzgerber:
Segment Extraction seems wrong. `range` is the containing object (chromatogram selection) and `item` is the contained object (noise segment) that should be included in the containing objects' range.


laeubi:
This should be fixed by https://github.com/eclipse/chemclipse/commit/43b76413c7e876d8587fc596558cd0e08ed45432 if the issue persists please open a new PR based on this commit as it contains also some other changes/improvements.

PR #139, opened 2019-11-21
Fix #129

### LINKED ISSUE #129: There should be an "intensity insensitive" mass spectrum comparator

kerner1000:
Change-Id: I5d24fe88a2915d9b789f689a0018c9b564e568d2
Signed-off-by: Alexander Kerner <alexander.kerner@lablicate.com>


laeubi (approved):
I think this suit our current needs, in the future I think we should make MassSpectrumComarators configurable then if 1/0 or the intensity is used can be controlled by a configuration option.

Issue #196, opened 2020-01-06
Peak Filter: Area%


eselmeister:
Add a filter to

- disable (set "Active for Analysis": false)
- delete

peaks, if certain area% constraints are matched (0-100%):

- Min Area%
- Max Area%

![demo-peak-filter](https://user-images.githubusercontent.com/686671/71830730-a7234680-30a7-11ea-9a1f-ee7b087198fb.png)

See peak filter examples in:

- org.eclipse.chemclipse.chromatogram.xxd.edit.supplier.snip.core.PeakFilter
- org.eclipse.chemclipse.xxd.model.filter.peaks.RemoveAllPeaksFilter
- ...

laeubi:
@alexstark4 is there anything left to do here?

PR #215, opened 2020-01-09
#195 #196 Implement peak filter for area and area% values

alexstark4:


alexstark4:
@eselmeister @laeubi @kerner1000 
Please have a look at this filter.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/ProcessPeaksByAreaFilter.java
```diff
@@ -0,0 +1,109 @@
...
+				compareAreaToLimit(compareAreaValue, areaLimits[1], option, listener, peak, false);
+			} else { // not within range, e.g. noisy peaks and big spikes
+				compareAreaToLimit(compareAreaValue, areaLimits[0], option, listener, peak, true);
+				compareAreaToLimit(compareAreaValue, areaLimits[1], option, listener, peak, false);
+			}
+			subMonitor.worked(1);
+		}
+	}
+
+	private <X extends IPeak> void compareAreaToLimit(double area, double limit, ProcessPeaksByAreaFilterTreatmentOption option, CRUDListener<X, IPeakModel> listener, X peak, boolean min) {
```
laeubi: Better use an comparator instead a boolean "min" that then decides how to compare

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/ProcessPeaksByAreaFilterSettings.java
```diff
@@ -0,0 +1,112 @@
...
+import org.eclipse.chemclipse.xxd.model.support.ProcessPeaksByAreaFilterAreaTypeSelection;
+import org.eclipse.chemclipse.xxd.model.support.ProcessPeaksByAreaFilterSelectionCriterion;
+import org.eclipse.chemclipse.xxd.model.support.ProcessPeaksByAreaFilterTreatmentOption;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+
+public class ProcessPeaksByAreaFilterSettings {
+
+	@JsonProperty(value = "Minimum area [integrated]", defaultValue = "5000000.0")
```
laeubi: Is this really a sensible default value?

laeubi (changes requested):
I think mixing Area percent and area in one filter is not a good idea as it makes the configuration more complex and the code hard to understand.
According to the issue https://github.com/eclipse/chemclipse/issues/195 there should be an option to either remove or disable the peaks, I can't find an option for this even though this might be better handled on an uper-level.

kerner1000 (changes requested):
I would also recommend to have two independent filters for that.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/ProcessPeaksByAreaFilterSettings.java
alexstark4: For my limited amount of example data the chosen values are reasonable in order to see the filters work.
Of course one can imagine that there can be other value ranges with matching sensible default values. Are you able to provide more data or insight into other value ranges or sensible default values?

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/ProcessPeaksByAreaFilter.java
alexstark4: A comparator is a comparison function, which imposes a total ordering on some collection of objects. But we do not (want to) pass anything to a sort method to allow precise control over the sort order.

Instead boolean min is used as a switch to distinguish between minimumValue and maximumValue for further calculations. It's not the boolean that is compared - the individual area values are compared.

Does that really have to be a comparator here or is the existing function sufficient?

BTW These parts have been re-factored completely.

laeubi: If it does not compare, why the method is named "compare...." then?

alexstark4:
> 
> 
> I think mixing Area percent and area in one filter is not a good idea as it makes the configuration more complex and the code hard to understand.
> According to the issue #195 there should be an option to either remove or disable the peaks, I can't find an option for this even though this might be better handled on an uper-level.

There are now 2 separate filters and the options mentioned in the issue are well implemented.

alexstark4:
> 
> 
> I would also recommend to have two independent filters for that.

There are now 2 separate filters.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/ProcessPeaksByIntegratedAreaFilterSettings.java
```diff
@@ -0,0 +1,73 @@
...
+import org.eclipse.chemclipse.support.settings.EnumSelectionSettingProperty;
+import org.eclipse.chemclipse.xxd.model.support.ProcessPeaksByAreaFilterSelectionCriterion;
+import org.eclipse.chemclipse.xxd.model.support.ProcessPeaksByAreaFilterTreatmentOption;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+
+public class ProcessPeaksByIntegratedAreaFilterSettings {
+
+	@JsonProperty(value = "Minimum area [integrated]:", defaultValue = "5000000.0")
```
kerner1000: Please remove default values. There is no defaults, that are meaningful in any case.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/support/ProcessPeaksByAreaFilterLocalSettings.java
```diff
@@ -0,0 +1,65 @@
...
+ * Alexander Stark - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.chemclipse.xxd.model.support;
+
+import org.eclipse.chemclipse.xxd.model.filter.peaks.ProcessPeaksByIntegratedAreaFilterSettings;
+import org.eclipse.chemclipse.xxd.model.filter.peaks.ProcessPeaksByPercentageAreaFilterSettings;
+
+public class ProcessPeaksByAreaFilterLocalSettings {
+
+	private double[] areaLimits = new double[] {0.0, 0.0};
```
kerner1000: Why is this an arrary? Please add javadoc to variables, where it is not perfectly clear what they mean.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/support/ProcessPeaksByAreaFilterUtils.java
```diff
@@ -14,43 +14,96 @@
...
+	public static <X extends IPeak> double calculatePercentageAreaCompareValue(X peak, Collection<X> read) {
+
+		return (100 / calculateAreaSum(read)) * peak.getIntegratedArea();
 	}
 
 	private static <X extends IPeak> double calculateAreaSum(Collection<X> read) {
-		
-		double areaSum = 0;		
+
+		double areaSum = 0;
```
kerner1000: For every peak, the area sum is here calculated over and over again.
I would rather calculate that sum once and re-use.
for the abs area filter, you dont need that code at all. try to get rid of the utils class and simplify the code a bit.

kerner1000 (changes requested):

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/support/ProcessPeaksByAreaFilterLocalSettings.java
alexstark4: In the meantime I re-factored this to
`private double lowerLimit = 0.0d;
private double upperLimit = 0.0d;`
 -> please refer to newer commits. I will be happy to add a Javadoc.

kerner1000 (approved):
Looks good to me.

Issue #197, opened 2020-01-06
Peak Filter: Leading/Tailing


eselmeister:
Add a filter to

- disable (set "Active for Analysis": false)
- delete

peaks, if certain leading/tailing constraints are matched:

- Min/Max Leading
- Min/Max Tailing

![demo-peak-filter](https://user-images.githubusercontent.com/686671/71830866-f79aa400-30a7-11ea-9240-a5253325c281.png)

See peak filter examples in:

- org.eclipse.chemclipse.chromatogram.xxd.edit.supplier.snip.core.PeakFilter
- org.eclipse.chemclipse.xxd.model.filter.peaks.RemoveAllPeaksFilter
- ...

laeubi:
@alexstark4 is there anything left to do here?

PR #221, opened 2020-01-21
Fix MANIFEST.MF last line(s) in xxd.model

alexstark4:
This will resolve the error 'Header must be terminated by a line break'
that was introduced while merging #195 #196 #197.


alexstark4:
@kerner1000 Please review this small commit.

Issue #228, opened 2020-01-23
Re-factor the latest filters


alexstark4:
Changes that did not make it into the pull requests  #215 and #218 :

- Shorten class names significantly

- separate PeakShapeFilter into two filters
1. Peak Shape Filter
2. Peak Assymetry Filter

- Shorten class names/descriptions e.g. "Leading value" -> "Leading"

- provide options for Treatment:
1. "Delete"
2. "Keep"
3. "Deactivate"
4. "Activate" 

- Shorten class names/descriptions for Criterion @ PeakShapeFilter 
1. "< Leading"
2. "> Tailing"
3. ">= Leading" and "<= Tailing" 

- add Preference Page(s)

PR #230, opened 2020-01-23
#228 Re-factor the latest filters

alexstark4:
* Shortened class names significantly
* split AsymmetricPeakShapeFilter into two filters
* Shortened class names/descriptions
* provided more options for Treatment
* Shortened class names/descriptions for Criterion


alexstark4:
TODO: add Preference Page(s)

@eselmeister @laeubi What is the most efficient method to implement Preference Page(s)? Alex K. suggested using Json to annotate the settings, but that doesn't create Preference Page(s).

alexstark4:
@kerner1000

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/PercentageAreaFilter.java
```diff
@@ -0,0 +1,144 @@
...
+		if(configuration == null) {
+			configuration = createConfiguration(read);
+		}
+		SubMonitor subMonitor = SubMonitor.convert(monitor, read.size());
+		double areaSum = calculateAreaSum(read);
+
+		for(X peak : read) {
+			double compareAreaValue = calculatePercentageAreaCompareValue(peak, areaSum);
+			applySelectedOptions(compareAreaValue, configuration, listener, peak);
+			checkForKeepOption(listener, configuration, peak);
```
kerner1000: Check the keep option inside the "processPeak" method.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/PercentageAreaFilter.java
```diff
@@ -0,0 +1,144 @@
...
+		switch (localSettings.getFilterTreatmentOption()) {
+		case ENABLE_PEAK:
+			peak.setActiveForAnalysis(true);
+			listener.updated(peak);
+			break;
+		case DEACTIVATE_PEAK:
+			peak.setActiveForAnalysis(false);
+			listener.updated(peak);
+			break;
+		case KEEP_PEAK:
```
kerner1000: I dont understand this. Why do you modify here the active for analysis property?
As far as I can tell, it should be:
if peak matches filter criterion, do nothing. Else, apply "Delete peak" option.

kerner1000 (changes requested):

kerner1000:
Next time please do changes such as refactoring of names in a dedicated commit.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/PercentageAreaFilter.java
alexstark4: The idea behind this is that the peak to be kept should be differentiated from the peak to be deleted. Therefore, the property is first modified and then the marked peaks are removed. In this way, the peaks sought are retained.

If I think about it again, this approach is not entirely optimal.

alexstark4: According to Philip, the requirement is that the peaks that are not kept should also be deleted.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/IntegratedAreaFilter.java
```diff
@@ -63,34 +63,60 @@ public boolean acceptsIPeaks(Collection<? extends IPeak> items) {
...
 			subMonitor.worked(1);
 		}
 	}
 
 	private static <X extends IPeak> void applySelectedOptions(double peakValue, IntegratedAreaFilterSettings configuration, CRUDListener<X, IPeakModel> listener, X peak) {
 
+		boolean keepFlag = false;
+		if(configuration.getFilterTreatmentOption()==ValueFilterTreatmentOption.KEEP_PEAK) {
+			keepFlag = true;
+		}
```
alexstark4: @kerner1000 Is it easier to understand? The flag is used to virtually invert the deletion of the peaks.

kerner1000:
See https://github.com/alexstark4/chemclipse/pull/1 for some suggestions.

kerner1000 (approved):

laeubi:
Please reopen with suitable ECA

PR #246, opened 2020-02-03
Refactor latest filters

alexstark4:
#228 Re-factor the latest filters - all in one commit ShapeFilter

* Opening a new PR on a new branch due to difficulties originating from merging, not successful rebasing etc.
* All commits from PR #230 are combined here in packages per filter.


chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/AreaFilter.java
```diff
@@ -0,0 +1,140 @@
...
+	@Override
+	public <X extends IPeak> void filterIPeaks(CRUDListener<X, IPeakModel> listener, AreaFilterSettings configuration, MessageConsumer messageConsumer, IProgressMonitor monitor) throws IllegalArgumentException {
+
+		Collection<X> read = listener.read();
+		if(configuration == null) {
+			configuration = createConfiguration(read);
+		}
+		SubMonitor subMonitor = SubMonitor.convert(monitor, read.size());
+		for(X peak : read) {
+			processPeak(listener, configuration, peak, getPredicate(configuration));
```
kerner1000: #getPredicate(configuration) Take this out of the loop, otherwise this is re-created for every peak.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/AsymmetryFilterSettings.java
```diff
@@ -0,0 +1,66 @@
...
+import org.eclipse.chemclipse.xxd.model.support.PeakAsymmetryFilterSelectionCriterion;
+import org.eclipse.chemclipse.xxd.model.support.ValueFilterTreatmentOption;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+
+public class AsymmetryFilterSettings {
+
+	@JsonProperty(value = "As:", defaultValue = "1.02")
+	@JsonPropertyDescription(value = "The peak asymmetry factor of a peak to be filtered accordingly.")
```
kerner1000, line 25: Add maybe, that this is tailing / leading. Maybe this factor could also be calculated by a strategy, so factor calculation is not fix but could be changed.
e.g. static final Supplier<Double> DefaultfactorSupplier = (p -> .getTailing() / getLeading());
Supplier<Doubl> factorSupplier = DefaultFactorSupplier;
getter/setter

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/ShapeFilter.java
```diff
@@ -0,0 +1,219 @@
...
+		}
+
+		public boolean test(double value) {
+
+			boolean result = predicate.test(value, shapeSetting);
+			return result;
+		}
+	}
+
+	private static class RangeContainer {
```
kerner1000, line 61: This could be replaced by Guava's 'Range' class.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/ShapeFilter.java
```diff
@@ -0,0 +1,219 @@
...
+import com.google.common.collect.Range;
+
+@Component(service = { IPeakFilter.class, Filter.class, Processor.class })
+public class ShapeFilter implements IPeakFilter<ShapeFilterSettings> {
+
+	private static BiPredicate<Double, Double> LEADING_SMALLER_THAN_LIMIT_COMPARATOR = (value, shapeSetting) -> (value < shapeSetting);
+	private static BiPredicate<Double, Double> TAILING_GREATER_THAN_LIMIT_COMPARATOR = (value, shapeSetting) -> (value > shapeSetting);
+	private static BiPredicate<RangeContainer, Range<Double>> VALUES_WITHIN_RANGE_COMPARATOR = (container, shapeSetting) -> (shapeSetting.contains(container.leading) && (shapeSetting.contains(container.tailing)));
+
+	private static class ShapePredicate<T> {
```
kerner1000, line 37: Try to use only one predicate, e.g.
MyPredicate<A,B>, where A and B can bei either Double or Range<Double>.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/ShapeFilter.java
```diff
@@ -0,0 +1,219 @@
...
+		case LEADING_SMALLER_THAN_LIMIT:
+			return new ShapePredicate<>(LEADING_SMALLER_THAN_LIMIT_COMPARATOR, configuration.getLeadingValue());
+		case TAILING_GREATER_THAN_LIMIT:
+			return new ShapePredicate<>(TAILING_GREATER_THAN_LIMIT_COMPARATOR, configuration.getTailingValue());
+		default:
+			throw new IllegalArgumentException("Unsupported Peak Filter Selection Criterion!");
+		}
+	}
+
+	private static <X extends IPeak> void processPeakConsideringRange(ShapeFilterSettings configuration, CRUDListener<X, IPeakModel> listener, X peak, RangePredicate<?> rangePredicate) {
```
kerner1000, line 157: Try to use only one method, process peak, using only one predicate, see above.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/support/ValueFilterTreatmentOption.java
```diff
@@ -13,26 +13,37 @@
...
-public enum ProcessPeaksByValueFilterTreatmentOption {
+/**
+ *  Describes how the filter will treat (e.g. process) the peak value with the
+ *  selected option:
+ *  <li>{@link #ENABLE_PEAK}</li>
+ *  <li>{@link #DEACTIVATE_PEAK}</li>
+ *  <li>{@link #KEEP_PEAK}</li>
+ *  <li>{@link #DELETE_PEAK}</li>
+ */
+public enum ValueFilterTreatmentOption {
```
kerner1000, line 24: Rename this to 'TreatmentOption'.

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/support/PeakAsymmetryFilterSelectionCriterion.java
```diff
@@ -0,0 +1,45 @@
...
+package org.eclipse.chemclipse.xxd.model.support;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ *  Describes the criterion to select the peak asymmetry factor:
+ *  <li>{@link #ASYMMETRY_FACTOR_SMALLER_THAN_LIMIT}</li>
+ *  <li>{@link #ASYMMETRY_FACTOR_GREATER_THAN_LIMIT}</li>
+ */
+public enum PeakAsymmetryFilterSelectionCriterion {
```
kerner1000, line 21: Rename this to 'SelectionOption'

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/support/AreaFilterSelectionCriterion.java
```diff
@@ -13,34 +13,31 @@
...
 
-public enum ProcessPeaksByValueFilterSelectionCriterion {
-
+/**
+ *  Describes the criterion to select the peak area:
+ *  <li>{@link #AREA_LESS_THAN_MINIMUM}</li>
+ *  <li>{@link #AREA_GREATER_THAN_MAXIMUM}</li>
+ *  <li>{@link #AREA_NOT_WITHIN_RANGE}</li>
+ */
+public enum AreaFilterSelectionCriterion {
```
kerner1000, line 22: Rename this

kerner1000 (changes requested):

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/AsymmetryFilterSettings.java
alexstark4: I will extend the description.

Can you explain why the factor should be calculated differently or changed later?

In our model we are calculating leading and tailing at 10% peak height. So according to IUPAC the factor is calculated as getTailing() / getLeading(). With each peak that has its own individual leading and tailing values.

If you want to change it later you should maybe consider to implement calculating leading and tailing at 5% peak height in addition to the existing way. Then it would be possible to calculate the factor according to USP, that is (getTailing() + getLeading()) / (2 * getLeading()).

Did you mean this kind of change or did you have something else in mind?

chemclipse/plugins/org.eclipse.chemclipse.xxd.model/src/org/eclipse/chemclipse/xxd/model/filter/peaks/ShapeFilter.java
alexstark4: Are you sure?

This is obviously an object to store two different peak property values that change for every inspected peak. We don't want to construct a range here because we are solely interested in these two values rather than a range.

alexstark4: Moreover, as stated in the Guava Range javadoc: _A range (or "interval") defines the boundaries around a contiguous span of values of some Comparable type; for example, "integers from 1 to 100 inclusive."_

That means we cannot construct/initiate half of the needed ranges because the leading is greater than tailing. Therefore I will stick to the RangeContainer but also try to use one predicate as proposed in the next comment.

alexstark4: grafik

However, this also means that if you want to filter according to the values ​​in the Peak/Scan List, the settings initialization will also fail.

kerner1000: correct, it is not a range.

PR #253, opened 2020-02-06
#228 Re-factor ShapeFilter and rename support classes

alexstark4:


kerner1000 (approved):
Looks good to me.

Issue #282, opened 2020-03-21
Extract/re-insert matrix from MSD chromatogram


lorenzgerber:
Currently, available MSD chromatogram data accessors (eg. extractedIonSignal) allow only extraction of objects where the abundances are stored per scan. 

It would be useful to have a direct method for the task of extracting and re-inserting the main data matrix (abundances per time(scan) and mz(ion)) of a MSD chromatogram.

Such a method would be beneficial for many signal processing algorithms that need direct access to either the signal along the time dimension or on the whole matrix at once.

laeubi:
@lorenzgerber  Thanks for contributing.

PR #283, opened 2020-03-21
#282 export/import matrix from msd chromatogram

lorenzgerber:


chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+	private float[] signal;
+	private boolean highRes = false;
+	
+	public ExtractedMatrix(IChromatogramSelectionMSD chromatogramSelection) {
+		this.selection = chromatogramSelection;
+		this.numberOfScans = selection.getStopScan() - selection.getStartScan() - 1;
+		this.scans = extractScans();
+		if(checkHighRes( 10)) {
+			highRes = true;
+			System.out.println("This is highres MS Data and should be binned first");
```
laeubi: Instead of system-out it would be better to throw an exception

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+		} else {
+			this.startIon = getMinMz();
+			this.stopIon = getMaxMz();
+			this.numberOfIons = this.stopIon - this.startIon + 1;
+			signal = new float[numberOfScans * (this.stopIon - this.startIon + 1)];
```
laeubi: the second part of the calculation is already stored in "numberOfIons"

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+			try {
+				for(int i = 0; i < numberOfScans; i++) {
+					for(int j = 0; j < scans.get(i).getIons().size(); j++) {
+						signal[ ((int) Math.round(scans.get(i).getIons().get(j).getIon() - this.startIon)) * numberOfScans + i    ] =  scans.get(i).getIons().get(j).getAbundance();
```
laeubi: why not using a "matrix" signal[][] instead? would make the code more readable and omit the calculation of indizies.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+			highRes = true;
+			System.out.println("This is highres MS Data and should be binned first");
+		} else {
+			this.startIon = getMinMz();
+			this.stopIon = getMaxMz();
+			this.numberOfIons = this.stopIon - this.startIon + 1;
+			signal = new float[numberOfScans * (this.stopIon - this.startIon + 1)];
+			try {
+				for(int i = 0; i < numberOfScans; i++) {
+					for(int j = 0; j < scans.get(i).getIons().size(); j++) {
```
laeubi: it would be better to extract

1. the scan (it is required in the inner loop anyways)
2. the size so it must not be retrieved on each iteration

beside that j could be renamed to 'ionIndex'

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+		if(checkHighRes( 10)) {
+			highRes = true;
+			System.out.println("This is highres MS Data and should be binned first");
+		} else {
+			this.startIon = getMinMz();
+			this.stopIon = getMaxMz();
+			this.numberOfIons = this.stopIon - this.startIon + 1;
+			signal = new float[numberOfScans * (this.stopIon - this.startIon + 1)];
+			try {
+				for(int i = 0; i < numberOfScans; i++) {
```
laeubi: Even though i is a common name  for a running index, in this case it might be better to name it 'scanIndex' instead

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+		IIonBounds bounds;
+		double rangeAbs;
+		IScanMSD currentScan;
+		
+		while(moreIonsThanFullMz == false) {
+			currentScan = iterator.next();
+			bounds = currentScan.getIonBounds();
+			rangeAbs = bounds.getHighestIon().getIon() - bounds.getLowestIon().getIon();
+			if(rangeAbs + limit < currentScan.getIons().size()) {
+				moreIonsThanFullMz = true;
```
laeubi: you can simply return true here

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+			}
+			if(!iterator.hasNext()) {
+				break;
+			}
+		}
+		
+		return moreIonsThanFullMz;
```
laeubi: just return false here

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+	}
+	
+	private Boolean checkHighRes( int limit ) {
+		boolean moreIonsThanFullMz = false;
+		Iterator<IScanMSD> iterator = scans.iterator();
+		IIonBounds bounds;
+		double rangeAbs;
+		IScanMSD currentScan;
+		
+		while(moreIonsThanFullMz == false) {
```
laeubi: instead of the while with iterator, you can use a simple enhanced for-loop (see comments below)

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+		int startRT;
+		int stopRT;
+		startRT = selection.getStartRetentionTime();
+		stopRT = selection.getStopRetentionTime();
+		scans = selection.getChromatogram() //
+				.getScans() //
+				.stream() //
+				.filter(s -> s instanceof IScanMSD) //
+				.map(IScanMSD.class::cast) //
+				.filter(s -> s.getRetentionTime() > startRT) //
```
laeubi: you probably want to use startScan/Stopscan instead of the RT? If using the RT, you probably want to include the outer borders by using >= ?

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+	}
+	
+	private int getMinMz() {
+
+		double min = scans.stream() //
+				.flatMap(scan -> scan.getIons().stream()) //
+				.min(Comparator.comparing(IIon::getIon)) //
+				.get() //
+				.getIon();
+		int minMz = (int)Math.round(min);
```
laeubi: You can use org.eclipse.chemclipse.msd.model.core.AbstractIon.getIon(double) instead, even though its functional equivalent its more descriptive and allows to change rounding sceme in a central place.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+	}
+	
+	private int getMaxMz() {
+
+		double max = scans.stream() //
+				.flatMap(scan -> scan.getIons().stream()) //
+				.max(Comparator.comparing(IIon::getIon)) //
+				.get() //
+				.getIon();
+		int maxMz = (int)Math.round(max);
```
laeubi: see above

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+		return (maxMz);
+	}
+
+	@Override
+	public float[] getMatrix() {
```
laeubi: as mentioned before, a matrix can better be expressed as float[][] in java.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+		if(highRes == true) {
```
laeubi: comparing with true is superfluous here as java is a typed language a boolean can only be true or false and there is no uncertainty

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+			return null;			
+		}
+		return this.signal;
+	}
+
+	@Override
+	public int getNumberOfScans() {
+		if(highRes == true) {
```
laeubi: see comment above

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+			return 0;			
+		}
+		return this.numberOfScans;
+	}
+
+	@Override
+	public int getNumberOfIons() {
+		if(highRes == true) {
```
laeubi: and so on ... :-)

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+		return this.numberOfIons;
+	}
+
+	@Override
+	public void updateSignal(float[] signal, int numberOfScans, int nubmerOfIons ) {
+		
+		IScanMSD currentScan;
+		IIon currentIon;
+		
+		if(highRes != true) {
```
laeubi: Simply use !highres

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+					currentScan.removeAllIons();
+					for(int j = startIon; j < stopIon; j++) {
+						if(signal[(j - startIon) * numberOfScans + (i-1)] != 0.0) {
+							currentIon = new Ion(j, signal[(j - startIon) * numberOfScans + (i - 1)] ); 
+							currentScan.addIon(currentIon);
+						}		
+					}
+				}
+			} catch (Exception e) {
+				e.printStackTrace();
```
laeubi: exceptions should not be suppressed, wrap it inside a runtime-exception and passing the original one as the cause if necessary.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,189 @@
...
+			}			
+		}
+	}
+
+	@Override
+	public int[] getScanNumbers() {
+		if(highRes == true) {
+			return null;			
+		}
+		int[] scanNumbers = scans.stream().mapToInt(scan -> scan.getScanNumber()).toArray();
```
laeubi, line 139: You could simply return an int[][] here to include scan number together with scan retentiontimes.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/IExtractedMatrix.java
```diff
@@ -0,0 +1,31 @@
...
+ * terms of the Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ * 
+ * Contributors:
+ * Lorenz Gerber - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.chemclipse.msd.model.matrix;
+
+
+public interface IExtractedMatrix {
```
laeubi: I think an interface is not needed here as its very unlikely that there are different implementations.
If not, the interface methods should contain at least a descriptive javadoc what these methods do and return.

laeubi (changes requested):
Please see the comments for suggested changes.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -17,39 +17,42 @@
...
-					for(int j = 0; j < scans.get(i).getIons().size(); j++) {
-						signal[ ((int) Math.round(scans.get(i).getIons().get(j).getIon() - this.startIon)) * numberOfScans + i    ] =  scans.get(i).getIons().get(j).getAbundance();
+				for(int scanIndex = 0; scanIndex < numberOfScans; scanIndex++) {
+					currentIons = scans.get(scanIndex).getIons();
+					numberIonsCurrentScan = currentIons.size(); 
+					for(int j = 0; j < numberIonsCurrentScan; j++) {
+						signal[ ((int) Math.round(currentIons.get(j).getIon() - this.startIon))] [j] =  currentIons.get(j).getAbundance();
 					}
 				}
 			} catch (Exception e) {
```
laeubi: You should either throw exceptions here to the caller or catch and rethrow as a spcialized exception describing the cause.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -59,25 +62,21 @@ public ExtractedMatrix(IChromatogramSelectionMSD chromatogramSelection) {
...
 	
 	private Boolean checkHighRes( int limit ) {
-		boolean moreIonsThanFullMz = false;
 		Iterator<IScanMSD> iterator = scans.iterator();
 		IIonBounds bounds;
 		double rangeAbs;
 		IScanMSD currentScan;
 		
-		while(moreIonsThanFullMz == false) {
+		while(iterator.hasNext()) {
```
laeubi: you can use foreach loop here

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -116,72 +115,49 @@ private int getMaxMz() {
...
+						currentScan.addIon(currentIon);
+					}		
 				}
-			} catch (Exception e) {
-				e.printStackTrace();
-			}			
-		}
+			}
+		} catch (Exception e) {
+			throw new RuntimeException(e);
```
laeubi: its a good idea to add a description here, at least something like "writing data back failed" as the exception message. In general its good not to simply catch "Exception" but narrow down the exception (types) as much as possible so you can give good hints to the user (e.g. "can't add ion")

laeubi (changes requested):
Some more changes would improve this from my point of view.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
lorenzgerber: try/catch actually not even needed here. Will remove.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,158 @@
...
+	private int startIon;
+	private int stopIon;
+	private float[][] signal;
+	
+	public ExtractedMatrix(IChromatogramSelectionMSD chromatogramSelection) {
+		this.selection = chromatogramSelection;
+		this.numberOfScans = selection.getStopScan() - selection.getStartScan() + 1;
+		this.scans = extractScans();
+		if(checkHighRes( 10)) {
+			throw new IllegalArgumentException();
```
laeubi: You should add a message here to the exception about the nature of problem e.g. "HighRes MSD is currently not supported"

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,158 @@
...
+			this.stopIon = getMaxMz();
+			this.numberOfIons = this.stopIon - this.startIon + 1;
+			signal = new float[numberOfScans][this.numberOfIons];
+			List<IIon> currentIons;
+			int numberIonsCurrentScan;
+			for(int scanIndex = 0; scanIndex < numberOfScans; scanIndex++) {
+				currentIons = scans.get(scanIndex).getIons();
+				numberIonsCurrentScan = currentIons.size(); 
+				for(int j = 0; j < numberIonsCurrentScan; j++) {
+					signal[ scanIndex ] [((int) Math.round(currentIons.get(j).getIon() - this.startIon))] =  currentIons.get(j).getAbundance();
```
laeubi: You should extract 'currentIons.get(j)' to prevent additional look-up, from the coode it looks as if you even don't need the for-loop and could convert to enhancedforloop as the array index is calculated from the ion and not the running-index.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -0,0 +1,158 @@
...
+				.stream() //
+				.filter(s -> s instanceof IScanMSD) //
+				.map(IScanMSD.class::cast) //
+				.filter(s -> s.getRetentionTime() >= startRT) //
+				.filter(s -> s.getRetentionTime() <= stopRT) //
+				.collect(Collectors.toList()); //
+		return (scans);
+	}
+	
+	private int getMinMz() {
```
laeubi: You could improve performance here by calculating min+max in one call see https://stackoverflow.com/questions/55213611/collect-both-min-and-max-in-one-stream for example.

Maybe you can even check if min/max could be calculated as part of "extractScans", because thats another point where all scans are iterated, so atm you interate the scanns three times.

laeubi (changes requested):

laeubi (approved):

Issue #289, opened 2020-03-25
interpolate signal abundances for ions with zero abundance


lorenzgerber:
Control software of some instruments allows setting minimum signal abundance threshold for recoding data. Abundances below this threshold are cut off. However, for some data/signal processing methods, it would be desirable to have a continuous signal.

A chromatogram filter/processor should be implemented that iterates over all ion channels and linear interpolates abundance values for ions with zero abundance.

laeubi:
Is this fixed now or are there any tasks left?

lorenzgerber:
Fixed for the moment

PR #290, opened 2020-03-25
#289 zero abundance ion interpolator draft

lorenzgerber:


eselmeister:
@lorenzgerber Was it intentionally, that the method "float[][] linearInterpolation(...)" is package protected instead of private?

Have you formatted the code using the ChemClipse code styles?

PR #293, opened 2020-03-26
Remove unused parameters and adjust dependent methods.

lorenzgerber:
In the current implementation of ExtractedMatrix, it is not possible to change the size of the matrix/number of scans, hence the numberOfIons, numberOfScans parameters are not used.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
@@ -24,7 +24,7 @@ import org.eclipse.chemclipse.msd.model.core.selection.IChromatogramSelectionMSD
 import org.eclipse.chemclipse.msd.model.implementation.Ion;
 
 public class ExtractedMatrix {
-	
+
 	private IChromatogramSelectionMSD selection;
 	private List<IScanMSD> scans;
 	private int numberOfScans;
@@ -32,12 +32,13 @@ public class ExtractedMatrix {
 	private int startIon;
 	private int stopIon;
 	private float[][] signal;
-	
+
 	public ExtractedMatrix(IChromatogramSelectionMSD chromatogramSelection) {
+
 		this.selection = chromatogramSelection;
 		this.numberOfScans = selection.getStopScan() - selection.getStartScan() + 1;
 		this.scans = extractScans();
-		if(checkHighRes( 10)) {
+		if(checkHighRes(10)) {
 			throw new IllegalArgumentException("HighRes MSD is currently not suported");
 		} else {
 			int[] minMaxMz = getMinMaxMz();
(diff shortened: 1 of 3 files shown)


chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -87,62 +87,63 @@ private Boolean checkHighRes( int limit ) {
...
 	}
 
 	public int getNumberOfIons() {
+
 		return this.numberOfIons;
 	}
 
-	public void updateSignal(float[][] signal, int numberOfScans, int nubmerOfIons ) {
-		
+	public void updateSignal(float[][] signal) {
```
laeubi: If you want going further, you can even omit the signal matrix here as it is already available in this.signal. At least you should not rely on the fact that the passed in array has the desired size (never trust the caller!)

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -87,62 +87,63 @@ private Boolean checkHighRes( int limit ) {
...
+		return (minMaxMz);
 	}
 
 	public float[][] getMatrix() {
+
 		return this.signal;
 	}
 
 	public int getNumberOfScans() {
+
```
laeubi: NumberOfscanns is the length of the matrix "this.signal" and must not be stored as an extra field

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
```diff
@@ -87,62 +87,63 @@ private Boolean checkHighRes( int limit ) {
...
 		return this.numberOfScans;
 	}
 
 	public int getNumberOfIons() {
+
```
laeubi: the number of ions is equal to the length of the first matrix element (this.signal[0].length) and thus could be ommited.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
lorenzgerber: removed method altogether and documented the dimensions of signal array in javadoc instead

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/matrix/ExtractedMatrix.java
lorenzgerber: removed method altogether and documented the dimensions of signal array in javadoc instead

laeubi (approved):

PR #504, opened 2021-03-02
Added feedback messages when invalid chromatogram selection ranges are set

### LINKED ISSUE #503: Select Range with invalid input resets the chromatogram selection
This is not really a bug, but a usability problem which chained with `ScanRemover` in a process method can cause total loss. When the detector was disabled in the first 5 minutes of retention time and a user selects 2 minutes the whole chromatogram is selected. In a batch process this causes less obvious followup errors which make it hard to spot the culprit. The same error at the end of the chromatogram is more intuitive as it just shows the whole chromatogram til the end even when a larger stop retention time is given.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/503


eselmeister:
@Mailaender Good. Please enclose the additional validations in curly brackets {}. Have you activated the code format settings?

Mailaender:
Added the optional brackets. That seems to be not enforced by the ChemClipse code style.

PR #531, opened 2021-03-18
Don't be alarmed when retention times are missing during database imports

### LINKED ISSUE #530: .msp database import with missing RI/MW spams the log file
Download https://github.com/MassBank/MassBank-data/releases/download/2021.03/MassBank_NIST.msp
Open it in OpenChrom.
Watch the console or look at ChemClipse.log which consists mostly of

> WARN  [ModalContext] org.eclipse.chemclipse.msd.converter.io.AbstractMassSpectraReader: java.lang.NumberFormatException: empty String

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/530


eselmeister:
@Mailaender Principally looks good. But it's the year in the header again. Additionally, I prefer squared brackets to enclose the if condition.

Mailaender:
Updated and added a source code comment.

PR #534, opened 2021-03-19
Added initial .mzXML support for MALDI data

Mailaender:
which is a popular open format due to partial vendor support.


eselmeister:
@Mailaender If you copy code from existing classes, then the copyright header must be remained.

Mailaender:
Fixed the headers.

Mailaender:
Fixed the ©️ years as well.

Issue #231, opened 2020-01-25
When a peak is deleted it remains visible in the chromatogram editor until next refresh/selection


laeubi:
Steps to reproduce:

- select a peak in peak/scan list
- delete it

alternatively "delete all peaks" filter can be used

PR #538, opened 2021-03-22
Fixed UI not updating when a peak was deleted from the Peak/Scans list

### LINKED ISSUE #60: A handful of little things
There are some small issues I'm confrontied with regularely.
I'm giving them in a bunch, hope that is ok.

1. **Updating the views** (tabs or windows) is annoying. Even when a tab is still activated (Task Quick Access), but covered by another, it holds the content from its last use. This sometimes leads to confusions and is a big obstacle for productive work flows. Every active tab should show the current data selected, it should be updated at least if it is brought to the screen after beeing covered by another tab.

2. Overview Mode provides a tab named **Miscellanious Info** that can be filled with text. Anyway, the text is not saved at all. I wanted to use this for documentation for what was done with the data. Since the tab **Edit History** is not accessable by the user and it is not filled automatically by applying an filter to the data, which I would prefere.

3. **Saving an OCB file** performes quite slowly. While loading takes 5 s, saving needs about 80 s and breaks entusiastic work experiences ;-)

4. The **Targets** tab is updated each time an identification (NIST) is performed, but the new results are added. I would say it should be refreshed by default. Because the old targets may not be correct if the peak was modified before, e.g. by scan subtraction or modification of single mass traces.

5. **New legend in the Chromatogram Overlay** tab is very usefull but does not support the SIC very well. The m/z values are given just at the end of the Description column and due to long description they are not visible. I would love to have them visible more directly, e.g. in a separate column. Additionally, it would be nice if the description could be modified via settings/preferences since it is not saved. Therefore, also the Data Name rather than the file name may be selected.

I think these issues would enhance the user experience immediately. So I hope you can find some time to work on some of these. Thanks a lot!

Mailaender:
Fixes https://github.com/OpenChrom/openchrom/issues/60 No. 4. This might be a regression as https://github.com/eclipse/chemclipse/issues/231 claims to have fixed this already.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedPeakScanListUI.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedPeakScanListUI.java
@@ -500,6 +500,9 @@ public class ExtendedPeakScanListUI extends Composite implements IExtendedPartUI
 			 */
 			if(scansToClear.size() > 0 || peaksToDelete.size() > 0) {
 				if(chromatogramSelection != null) {
+					chromatogramSelection.setSelectedPeak(null);
+					chromatogramSelection.setSelectedScan(null);
+					chromatogramSelection.setSelectedIdentifiedScan(null);
 					chromatogramSelection.update(true);
 				}
 			}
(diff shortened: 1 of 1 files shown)


eselmeister:
@Mailaender It's ok to use "TOPIC_IDENTIFICATION_TARGETS_UPDATE_SELECTION" here. A String argument is used, hence the ExtendedTargetsUI assigns null to its internal object and thus resets the table. The event system has been improved in autumn 2020 and we got rid of some events of the early days of ChemClipse/OpenChrom. When closing an editor, the event "TOPIC_EDITOR_CHROMATOGRAM_CLOSE" is used and processed. I would recommend to find a similar strategy for Scan, Peaks and Targets. Before we introduce new events, I'd like to create a sketch of existing events and I'd like to develop a consistent strategy how to use events in ChemClipse/OpenChrom.

PR #543, opened 2021-03-23
Reduced unnecessary UI updates

### LINKED ISSUE #540: Chromatogram right-click menu immediately closes
I believe this happens because the right-click triggers an update (as seen in the console) of SWT Chart which then kills the menu. If you left-click on the chart first, wait a few seconds and then right-click the menu won't immediately close.

Mailaender:
Fixes https://github.com/eclipse/chemclipse/issues/540


eselmeister:
@Mailaender The change in the chromatgram selection looks fine. The only issue I've found is, that the scan targets part and chart/table is not updated when switching to a reference chromatogram.

Bildschirmfoto von 2021-03-24 16-48-34

As this change leads to a better usability of the chromatogram popup menu, it's IHMO acceptable.

But I would strongly recommend to skip the cache handling in the TargetsPart, as the TargetSupplier could be the same but some target labels might have been modified or e.g. the mol weight was calculated for each entry via the chromatogram menu. It might be confusing if the changes in the targets entries are not displayed.

Bildschirmfoto von 2021-03-24 17-14-26

It should be updated after running a filter to:

Bildschirmfoto von 2021-03-24 17-14-34

Mailaender:
I think I finally identified and removed the real culprit. https://github.com/eclipse/chemclipse/pull/543/commits/88ee2c5c27842f13fe27ab2b27404926e2813971 A lot of the UI updates indeed seem to rely on this spamming events even when the part is empty. I added updates for scan table/chart similar to the peak list which already behaves properly. I also don't cache the individual targets anymore only the chromatogram. I haven't tested the mol weight calculation as my current setup in Eclipse is either ChemClipse or OpenChrom not mixed together.

PR #545, opened 2021-03-24
Fixed WindowSize enum being compared with an integer

Mailaender:
In Java enums are not casted to integer automatically. See https://docs.oracle.com/javase/8/docs/api/java/lang/Enum.html#ordinal--


eselmeister:
@Mailaender Please adjust the year in the copyright header.

Mailaender:
Sorry, fixed.

PR #568, opened 2021-04-15
Fixed logging InvocationTargetException without cause

Mailaender:
This will give more meaningful stacktraces and log entries.


eselmeister:
@Mailaender Don't forget to update the year in the copyright header.

Mailaender:
Updated.

PR #570, opened 2021-04-15
Overhauled the UI preferences titles

Mailaender:
Avoids redundancy like
> Settings > Settings > Chromatogram Settings

and shortens the title as the preferences dialog has a small width.


eselmeister:
@Mailaender Don't forget to update the year in the copyright header.

Mailaender:
Updated.

PR #586, opened 2021-05-05
Added a reset perspective function

### LINKED ISSUE #585: Reset Perspective
While the RCP UI is super flexible novice user can get lost easily when accidentally dragging or closing windows. The Eclipse IDE has a `Windows > Reset Perspectives` function which quickly goes back to a default layout without data loss.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/585.

For this to work properly each perspective requires changes to the `.e4xmi` 
* with placeholder ID matching the part definition because that is what gets restored
* part needs to be defined in the same project/file. I couldn't get imports of shared elements from the platform to work.


eselmeister:
@Mailaender Removing the flag "activatePartsInitially" may lead to a reset of the "Data Analysis" perspective, whenever the user switches to it. Thinks of the following workflow:

A) Select the "Data Analysis" perspective and activate/deactivate manually some parts
B) Switch to the "PCA" perspective and make some analysis
C) Switch back to the "Data Analysis"

What happens?
I assume, that the "Data Analysis"  is reset to its initial status. That's probably something, which could be confusing.
Could you test and evaluate this case?

Mailaender:
You were right. I added a new event to fix that.

PR #615, opened 2021-05-26
Fix #614 allow Classifiers to work also in WSD

### LINKED ISSUE #614: Deprecated AbstractChromatogramClassifier constructor defaults to MSD datatype only
The deprecated AbstractChromatogramClassifier constructor defaults to DataType.MSD which prevents classifiers that extend AbstractChromatogramClassifier and work both in MSD and WSD from showing up in WSD chromatograms.

lorenzgerber:


chemclipse/plugins/org.eclipse.chemclipse.chromatogram.msd.classifier/src/org/eclipse/chemclipse/chromatogram/msd/classifier/core/AbstractChromatogramClassifier.java
```diff
@@ -27,10 +28,12 @@
 
 	@Deprecated
 	public AbstractChromatogramClassifier() {
-		this(DataType.MSD);
+
+		this(DataType.MSD, DataType.WSD);
```
laeubi, line 32: instead of calling this constructor (what now claims that MSD and WSD is supported for **all** derived types, derived types should simply not call` super()`  constructor (as it is deprecated anyways) but instead call `super(DataType.MSD)` or `super(DataType.WSD)` ... (or even multiple types if desired)

laeubi (changes requested):

eselmeister:
Done, solved otherwise.

PR #641, opened 2021-06-20
Added the MALDI-TOF MS perspective

Mailaender:
from @OpenChrom to avoid redundancy. This is also connected to https://github.com/eclipse/chemclipse/pull/534 which makes this fully featured and standalone.


eselmeister:
@Mailaender Looks good. But OpenChrom has become a versatile tool to analyze/visualize data also from other techniques like PCR, FTIR and NMR.

PR #655, opened 2021-06-24
Added a stopwatch to the magic number checks

Mailaender:
to investigate https://github.com/OpenChrom/openchrom/issues/101.


eselmeister:
It looks good. As logging also consumes processor time and space for the log file on disk, we could probably add a system option to enable/disable the MagicNumberMatcher logging on demand.

Mailaender:
Yes, I was a bit torn on it as a check against the preferences will also cost CPU (haven't profiled any of it) so maybe increasing `if(spent > 0)` miliseconds might be a way. As I don't have any clue whether those performance problems are caused by large spikes from a single unoptimized converter or just the shere number of magic number checks I kept it track almost everything.

eselmeister:
Yep, good idea. Let's keep it so far as it is. If we discover performance issues due to the stopwatch, we could have a look at this again. There are several ways, how to address it:

- increase the "spent time" limit for reporting
- Add a VM argument -Denable.magicnumber.stopwatch=true
- ...

PR #662, opened 2021-06-28
Added a configuration option on how to label PCR data

### LINKED ISSUE #661: PCR data labeling misses the vital information
Currently, PCR data is labeled by well number and internal channel number. The 96 plates are usually addressed by the coordinate. More importantly is the sample name. Some people may even prefer both, so it should be configurable.

assay plate

The channel is static for the plate chart, so it is not of much use. The user-friendly way is to display the channel wavelengths and not a numerical enumerator.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/661

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/preferences/PreferenceConstants.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/preferences/PreferenceConstants.java
@@ -15,6 +15,7 @@ package org.eclipse.chemclipse.ux.extension.xxd.ui.preferences;
 import org.eclipse.chemclipse.model.core.support.HeaderField;
 import org.eclipse.chemclipse.model.traces.NamedTraceUtil;
 import org.eclipse.chemclipse.msd.model.support.CalculationType;
+import org.eclipse.chemclipse.pcr.model.core.support.LabelSetting;
 import org.eclipse.chemclipse.swt.ui.support.Colors;
 import org.eclipse.chemclipse.ux.extension.ui.support.PartSupport;
 import org.eclipse.swt.SWT;
@@ -780,6 +781,8 @@ public class PreferenceConstants extends ChartOptions {
 	public static final String DEF_PCR_COLOR_CODES = "";
 	public static final String P_PCR_SAVE_AS_FOLDER = "pcrSaveAsFolder";
 	public static final String DEF_PCR_SAVE_AS_FOLDER = "";
+	public static final String P_PCR_REFERENCE_LABEL = "pcrReferenceLabel";
+	public static final String DEF_PCR_REFERENCE_LABEL = LabelSetting.COORDINATE_SAMPLENAME.name();
 	/*
 	 * Processor
 	 */
--- a/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedPlateChartsUI.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedPlateChartsUI.java
@@ -159,8 +159,8 @@ public class ExtendedPlateChartsUI extends Composite implements IExtendedPartUI 
 		buttonToolbarInfo = createButtonToggleToolbar(composite, toolbarInfo, IMAGE_INFO, TOOLTIP_INFO);
 		createButtonToggleChartLegend(composite, chartControl, IMAGE_LEGEND);
 		createResetButton(composite);
-		createSettingsButton(composite);
 		createColorCompensationButton(composite);
+		createSettingsButton(composite);
 	}
 
 	private void createSettingsButton(Composite parent) {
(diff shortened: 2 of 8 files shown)


chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedPlateChartsUI.java
```diff
@@ -271,6 +273,18 @@ private void updateChart() {
 		}
 	}
 
+	private String getLabel(IWell well) {
+
+		LabelSetting referenceLabel = LabelSetting.valueOf(preferenceStore.getString(PreferenceConstants.P_PCR_REFERENCE_LABEL));
+		if(referenceLabel.equals(LabelSetting.SAMPLENAME))
```
eselmeister: Please add curly braces. Are the formatting settings set in your IDE?

chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/preferences/PreferenceConstants.java
```diff
@@ -778,6 +779,8 @@
 	public static final String DEF_PCR_DEFAULT_COLOR = "192,192,192";
 	public static final String P_PCR_COLOR_CODES = "pcrColorCodes";
 	public static final String DEF_PCR_COLOR_CODES = "";
+	public static final String P_PCR_REFERENCE_LABEL = "pcrReferenceLabel";
+	public static final String DEF_PCR_REFERENCE_LABEL = LabelSetting.COORDINATE_SAMPLENAME.getLabel();
```
eselmeister: I would recommend to persist "LabelSetting.COORDINATE_SAMPLENAME.name()". The label itself if variable and could be translated in the future. The enum entry name itself is independent of a possibly translated label.

chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedPlateChartsUI.java
```diff
@@ -271,6 +273,18 @@ private void updateChart() {
 		}
 	}
 
+	private String getLabel(IWell well) {
+
+		LabelSetting referenceLabel = LabelSetting.valueOf(preferenceStore.getString(PreferenceConstants.P_PCR_REFERENCE_LABEL));
```
eselmeister: Handle: IllegalArgumentException | NullPointerException

chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedPlateChartsUI.java
Mailaender: I enabled the `Use blocks in if/while/for/do statements` save action now.

chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedPlateChartsUI.java
Mailaender: I don't see them handled anywhere, so I will just add a default fallback here.

Mailaender:
Updated.

eselmeister:
What happens if 

`LabelSetting referenceLabel = LabelSetting.valueOf(preferenceStore.getString(PreferenceConstants.P_PCR_REFERENCE_LABEL));`

the return of `preferenceStore.getString(PreferenceConstants.P_PCR_REFERENCE_LABEL)` is null or "" or doesn't match any enum name?

Mailaender:
```java
	private String getString(Properties p, String name) {
		String value = p != null ? p.getProperty(name) : null;
		if (value == null) {
			return STRING_DEFAULT_DEFAULT;
		}
		return value;
	}
```

will never return null. `STRING_DEFAULT_DEFAULT` is an empty string. I added a fallback to the default value instead of no label. Also added enum parsing that will not crash the widget, but also go to a fallback.

I now also set the ID of line series and use the description field instead.

eselmeister:
Looks good. Two comments:

A) When introducing a new dependency, I recommend to also set the required version in the MANIFEST.MF:
`com.google.guava;bundle-version="27.1.0"`

B) I recommend to use as less dependencies as possible. In this case, Google Enums is not necessarily needed. You could rewrite the code in the following way, which is also more compact:

```
private String getLabel(IWell well) {

	LabelSetting labelSetting = getLabelSetting();
	switch(labelSetting) {
		case SAMPLENAME:
			return well.getSampleId();
		case COORDINATE:
			return well.getPosition().toString();
		case COORDINATE_SAMPLENAME:
		default:
			return well.getPosition().toString() + ": " + well.getSampleId();
	}
}

private LabelSetting getLabelSetting() {

	try {
		return LabelSetting.valueOf(preferenceStore.getString(PreferenceConstants.P_PCR_REFERENCE_LABEL));
	} catch(Exception e) {
		return LabelSetting.COORDINATE_SAMPLENAME;
	}
}

```

Mailaender:
Rebased, removed Google Guave, initialized the setting and put the settings button on the right.

PR #669, opened 2021-06-29
Removed ProteoMS

### LINKED ISSUE #668: Drop ProteoMS
I experimented with the `Proteoms` perspective, which is not included by default. It contains project management and a redundant `.mgf` file parser which however does not recognize files from https://zenodo.org/search?page=1&size=20&file_type=mgf

image

I only got
> ERROR main importdata.ImportWizard.performFinish (ImportWizard.java:112) Import error. null

The UI is full of placeholders in hello world style. 

https://github.com/eclipse/chemclipse/blob/9921848d8970bcb996d71ce9a416d06437026c49/chemclipse/plugins/org.eclipse.chemclipse.msd.identifier.supplier.proteoms.ui/src/org/eclipse/chemclipse/msd/identifier/supplier/proteoms/ui/parts/ProteomPeptideAnalyzerPart.java#L36

It contains its own model for primitives giving existing terminology a different meaning instead of using the ChemClipse ones:

https://github.com/eclipse/chemclipse/blob/9921848d8970bcb996d71ce9a416d06437026c49/chemclipse/plugins/org.eclipse.chemclipse.msd.identifier.supplier.proteoms/src/org/eclipse/chemclipse/msd/identifier/supplier/proteoms/model/Peak.java#L18-L20

It seems to be abandoned in early stages with nothing left to salvage. I vote for removal.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/668


eselmeister:
You are right, this project was a promising test, but it didn't work out.

PR #691, opened 2021-08-04
Removed trademarked icons

Mailaender:
It seems they are outdated as neither ChemClipse nor @OpenChrom seem to use them. Still the different IPs should not be mixed.


eselmeister:
@Mailaender These are old OpenChrom icons. They are residuals from the days, before large parts of OpenChrom were migrated to the Eclipse Foundation. Good point to remove them.

But please keep the folders:
7x8 (used for decorators)
75x66

As they are referenced in the interface "IApplicationImageProvider".

Mailaender:
I kept the empty folders now.

PR #703, opened 2021-08-10
Fixed editor being marked as dirty when there are no changes

### LINKED ISSUE #702: Opened files are immediately set as changed
https://www.eclipse.org/lists/chemclipse-dev/msg00068.html mentioned this before as it irritates every new user. Opening a file and closing it will always cause a nag dialog to save non-existing changes, which is annoying and also unexpected. This seems to be a quick hack to force enable the save button. However, simply clicking on the open editor already enables save as.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/702


eselmeister:
Copyright header: please adjust the year to 2021.

Mailaender:
Updated.

PR #704, opened 2021-08-11
Removed notes directed at developers from end user logs

Mailaender:
https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners is a better solution to enforce this policy. The log is already quite crowded, which makes it hard to find actual problems in it.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.xxd.converter.supplier.chemclipse/src/org/eclipse/chemclipse/xxd/converter/supplier/chemclipse/Activator.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.xxd.converter.supplier.chemclipse/src/org/eclipse/chemclipse/xxd/converter/supplier/chemclipse/Activator.java
@@ -11,13 +11,11 @@
  *******************************************************************************/
 package org.eclipse.chemclipse.xxd.converter.supplier.chemclipse;
 
-import org.eclipse.chemclipse.logging.core.Logger;
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.BundleContext;
 
 public class Activator implements BundleActivator {
 
-	private static final Logger logger = Logger.getLogger(Activator.class);
 	private static BundleContext context;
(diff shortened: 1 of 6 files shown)


chemclipse/plugins/org.eclipse.chemclipse.xxd.converter.supplier.chemclipse/src/org/eclipse/chemclipse/xxd/converter/supplier/chemclipse/Activator.java
```diff
@@ -29,30 +27,17 @@ public static BundleContext getContext() {
 	 * (non-Javadoc)
 	 * @see org.osgi.framework.BundleActivator#start(org.osgi.framework.BundleContext)
 	 */
+	@Override
 	public void start(BundleContext bundleContext) throws Exception {
 
 		Activator.context = bundleContext;
-		//
```
eselmeister, line 35: I would like to keep this part as it shows up in the console when testing the application. It's a strong hint for the developers to take care of the *.ocb format. As we can't enforce more strict permissions on the *.ocb bundle, the message shall ensure that no modification happens that is not reviewed and authorized.

Mailaender: I debug a lot via the console. To me, it is just visual clutter and noise. It is displayed and scrolled away fast. I doubt there will be a sudden influx of developers making changes and even if so GitHub code owners protection does it way more effectively.

eselmeister: OK, let's try it with the CODEOWNERS directive.

Issue #694, opened 2021-08-09
Chromatogram Overlay - XWC is slow


eselmeister:
The selection of XWC (Extracted Wavelength Chromatogram) in the chromatogram overlay leads to performance issues when trying to display HPLC-DAD chromatograms.

eselmeister:
The optimized modus is used by default now. If one likes to display all wavelength of a HPLC-DAD chromatogram, the optimized option needs to be unchecked. But then it takes a while to extract and display the complete data set.

![Bildschirmfoto von 2021-08-09 10-49-27](https://user-images.githubusercontent.com/686671/128689659-cd181d75-290e-4c83-afc8-cc7ee31f0286.png)

PR #706, opened 2021-08-11
Added a Max (Spectrum) Plot Chromatogram

### LINKED ISSUE #705: Max Plot Chromatogram
A Max Plot Chromatogram is a visualization used in HPLC-DAD method development where each point gets plotted at maximum absorbance. This mimics a scenario where each peak is optimized for quantification at a specific wavelength, but displayed in a combined view. It is not to be used for quantification itself, as it is an artificial appearance.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/705. Thanks for fixing https://github.com/eclipse/chemclipse/issues/694 as that really helped creating and verifying this.


eselmeister:
Looks good so far. Please adjust the copyright header to 2021 and append your name in the list of contributors.

Mailaender:
Updated.

PR #724, opened 2021-08-27
Reduced warnings

Mailaender:
Only the preference initializers should fix real breakage. There is possibly more hidden under all the garbage.


eselmeister:
Looks good. Two issues:

- ExtendedPeakChartUI - adjust the year in the header
- getBackground() changes

The method `public float getBackgroundAbundance(int retentionTime)` must be not marked as deprecated. I can't recall from the history, why it was marked with the `@Deprecated` annotation. Please remove the `@Deprecated` annotation from the method and skip the changes `getBackgroundAbundance(...)` -> `getBackground(...)`. The method getBackground(...) may return Float.NaN if `defaultBackgroundAbundance` is specified in the constructor as NaN. This could lead to exceptions when `dataOutputStream.writeFloat(...)` is written.

Mailaender:
I understand. Updated.

I leave the `getBackgroundAbundance()` → `getBackground()` case to you.

There is a similar case of a not properly deprecated API with `getPeak()` → `getPeaks().get()` which however still has unit tests that explicitly test the criticized behavior of said method so I also don't know how to proceed there.

https://github.com/eclipse/chemclipse/blob/9921848d8970bcb996d71ce9a416d06437026c49/chemclipse/plugins/org.eclipse.chemclipse.model/src/org/eclipse/chemclipse/model/core/IPeaks.java#L56-L73

eselmeister:
Yep, I'll take care of it.

PR #729, opened 2021-09-03
Use plural forms in the PCA welcome tile

Mailaender:
as it is not possible to run PCA with just one chromatogram.


eselmeister:
Good point. But the year in the copyright header needs to be adjusted to 2021.

Mailaender:
Updated.

PR #737, opened 2021-09-14
Replaced window size enum with validated integer

### LINKED ISSUE #736: Window Size is stored and displayed as an enum
Both is wrong imho it is irritating to the user and the developer a like. This is technical dept from a time when settings couldn't be validated properly. In fact, now everything is in place to check if user input is an uneven range of integers. As old process methods should just still work, there has be some kind of backwards compatibility feature to translate old settings.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/736


eselmeister:
Are settings validated correctly, that have been persisted already? E.g., if a setting "WIDTH_5" exists already, it needs to be translated to 5, otherwise the process might crash as the existing setting can't be parsed.

Mailaender:
It took me multiple attempts to get backwards compatibility. I went with `@JsonDeserialize(using = WindowSizeDeserializer.class)` and I exploit the deserialization to always give me a number I can work with. Try using an old process method. It will still run even with legacy settings. Open it up and the editor will tell you to change the value. So this should be a soft update.

eselmeister:
Does it also work via the Eclipse preference page? In recent time, we have sometimes encountered some problems when underlying values/strategies have changed. Could you create some test cases to ensure the backward compatibility?

Mailaender:
Alright, this is good to go from my side now. I added a preferences widget that will convert old values on the fly. Process methods will now also work flawlessly and auto-convert old values both during execution and in the UI settings widget. There was a subtle, but hard to find and fix bug in the convoluted settings to UI mapping. It now behaves just like the deserialized process function, in a way that every object has the correct type and custom deserialisation attached to it. It was hard-coded to string conversions before for no apparent reason.

eselmeister:
Looks good. Some years in the copyright header need adjustments, but I'll take of it this time.

PR #750, opened 2021-09-27
Fixed the PCA perspective from freezing during initialisation

### LINKED ISSUE #749: PCA perspective freezes on startup
This can be tracked down to icon initialization. The caching is subtly wrong. It just hangs the main thread instead of failing properly.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/749


Mailaender:
I believe this may have been a race condition between the cache generation and the draw function, which may be on a different thread. Doing this lookup with every draw is probably also wasteful, so this might be a small performance win as well.

eselmeister:
The PCA perspective still freezes, so I assume it is another problem. Anyways, the change looks reasonable. But please adjust the year in the copyright header.

Mailaender:
I changed it back to my previous hacky solution, which seems to fix the problem and not the cleanups.

PR #754, opened 2021-09-29
Reduced the info level warnings

Mailaender:
It seems that JUnit improved, since a lot of those unit tests were written. Some were using comparisons that were impossible due to Java's built-in type safety which issued warnings in the IDE, others used double negation and most of them only exposed a boolean value to the logs when actually objects were compared which limits the error messages they give. In the end, nothing really changes. Getting rid of those warnings revealed one real problem.


eselmeister:
Looks good. Please adjust header to 2021 in:

- PeakQuantifierSupplier_3_Test
- PeakQuantifierSupplier_4_Test
- TotalIonSignal_4_Test
- TotalIonSignal_5_Test

Mailaender:
Updated.

PR #756, opened 2021-09-30
Removed ALL CAPS settings from the 1st derivative peak detection settings dialog

### LINKED ISSUE #755: CAPS in the Settings User Interface
Java enums are `SCREAMING_SNAKE_CASE` which seems to be best practice and it helps me personally with readability of the source code. However, those raw enums are also displayed in the UI which I find pretty terrible. Some also have bad grammar in conjunction with the description. Oftentimes, it is clear what is meant by it, but I think it is bad polish. Combo boxes should differentiate between labels and raw data. The goal is to not touch deserialization or default values to avoid any problems with backwards compatibility. It should just look pretty.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/755


eselmeister:
@Mailaender Please have a look at the conflicting serialization file.

Mailaender:
Rebased. This had to happen as I applied https://github.com/eclipse/chemclipse/commit/767c1519348d9caa64a94473f89a873f9710a3e9 here as well.

PR #774, opened 2021-10-11
Use the new ILabel interface for automatic combo box labelling

Mailaender:
This is a followup of https://github.com/eclipse/chemclipse/pull/756.


eselmeister:
Looks good. But please keep the original copyright header instead of replacing them by "ChemClipse Contributors". Set yourself to the contributors list.

Mailaender:
Updated.

PR #781, opened 2021-10-13
Updated peak max and trapezoid integrator settings

### LINKED ISSUE #571: Peak Integrator Trapezoid preference page settings are ignored
image

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/571


eselmeister:
Looks good. The integrator area constraints have been implemented in the beginning of OpenChrom. It was inspired by the ChemStation, as the ChemStation offers integration options like "Area Reject", "Min Peak Width", ... . These options are still available for the integrator. Nowadays, we have better options to filter peaks. But we should keep the basic functionality, even if they are not used. IMHO, removing the settings from the preferences is ok, as they never have been used.

PR #784, opened 2021-10-15
Fixed data not unloading properly when the parent editor was closed

### LINKED ISSUE #599: Closing the PCA editor does not unload data
Even a small data set accounts for ~1 GB of RAM.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/599


chemclipse/plugins/org.eclipse.chemclipse.chromatogram.xxd.process.supplier.pca.ui/src/org/eclipse/chemclipse/chromatogram/xxd/process/supplier/pca/ui/editors/PCAEditor.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2020 Lablicate GmHb.
+ * Copyright (c) 2018, 2021 Lablicate GmHb.
```
Mailaender, line 2: Typo

eselmeister:
The `plot.deleteSeries();` method call in `void updatePlot(int pcX, int pcY)` is not necessarily required as it's called in the `setInput` method. The method `setInput` is called in any case, hence `deleteSeries()` is also called even if the data is null.

Mailaender:
Yes, that was one of my earlier attempts to fix it. Maybe I should have removed them.

PR #804, opened 2021-11-01
Fixed a regression in column type labeling

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/803

PR #808, opened 2021-11-16
Revert column labelling

Mailaender:
This reverts https://github.com/eclipse/chemclipse/pull/774 and https://github.com/eclipse/chemclipse/pull/804. The backslash of regressions does not justify the minor polish win. Sorry.


eselmeister:
I would not recommend to just revert the commit. I found a way to improve the column handling by keeping the backward compatibility. Let me work on it, see
https://github.com/eclipse/chemclipse/issues/816

PR #833, opened 2021-12-02
Fixed NumberFormatExceptions in TableSupport

Mailaender:
This is the same as https://github.com/eclipse/swtchart/pull/277/ avoid logging warnings that are irrelevant. Exception handling is also a performance drain.


eselmeister:
Even if valid, I would recommend to put if conditions in curly brackets.

Mailaender:
Sorry, I didn't have https://stackoverflow.com/a/3704350 enabled.

PR #852, opened 2021-12-16
Fixed mzML array length compatiblity problems

Mailaender:
Resolves the issues raised in https://github.com/OpenChrom/openchrom/issues/175#issuecomment-995318249


bspratt:
Can you send me your latest conversion output for testing? Thanks. I feel we are very close to making this work (though from my reading of your changes it looks like maybe you are not setting the defaultArrayLength value?).

Mailaender:
Sorry, I misunderstood and thought the example file was generated using OpenChrom and that was already set.

PR #854, opened 2021-12-16
Remove unused log4j 2.15.0

Mailaender:
resolves CVE-2021-45046. It is not used. Distributing this only makes it look like ChemClipse derived software is vulnerable. We don't need any of the remote configuration loading and class instantiation features of version 2. This reverts https://github.com/eclipse/chemclipse/commit/cd625cd025b288afd4d889f405dcd47939af6331


Mailaender:
This however will not remove the much older log4j2 in the current orbit which this will override.

PR #858, opened 2022-01-04
Fixed HPLC-DAD .ocb peak import

### LINKED ISSUE #857: HPLC-DAD files don't save peaks
This is probably a regression from https://github.com/eclipse/chemclipse/pull/776.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/857


eselmeister:
Yep, that could be a fix for the problem. But

`peakMaximum.addScanSignal(new ScanSignalWSD(0, totalSignalScan));`

sets a peak without any wavelength information. A peak could consist of several wavelengths when using HPLC-DAD. In case of HPLC-UV/Vis, most commonly only one wavelength at a time is used. Probably, we should create a new version soon, that also allows to store n wavelengths for a WSD peak. Additionally, the adjustTotalIntensity might need improvements. In case of GC-MS it's easy as only positive values occur. In case GC-FID, we have only one signal. In case of HPLC-UV/Vis too. But with HPLC-DAD, positive and negative signals might be available.

Give me some time to think about side-effects regarding this PR. I'll come to a decision today.

Mailaender:
I was inspired by 
https://github.com/eclipse/chemclipse/blob/9921848d8970bcb996d71ce9a416d06437026c49/chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/core/IIon.java#L45

eselmeister:
Yep, that was also the right choice. In a next version, we should improve the HPLC-DAD peak handling. Let's chat.

PR #875, opened 2022-01-20
Set chromatograms as dirty on lower levels

### LINKED ISSUE #874: Chromatogram Selection (Select Range) is broken
It resets immediately after being set.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/874. Followup of https://github.com/eclipse/chemclipse/pull/873 and https://github.com/eclipse/chemclipse/pull/730.


eselmeister:
Looks good.

PR #877, opened 2022-01-20
Fixed database editor not being marked as dirty properly

Mailaender:
This is a followup of https://github.com/eclipse/chemclipse/pull/876.


eselmeister:
For better readability, I recommend to list the methods by access option:

- public
- protected
- private

Mailaender:
There is even a button for it https://wiki.eclipse.org/Recommenders/CleanCodeMethodSorter

PR #886, opened 2022-01-27
Added undo/redo for peak/scan deletion

### LINKED ISSUE #887: Undo/Redo buttons never enable
The feature is completely disabled.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/887


eselmeister:
It's a huge change. But I agree, that the do/redo/undo for chromatogram instances was too heavy. Each and every time, a complete dump of the chromatograms was created, which ensured the consistency of the data, but was slow. Hence, it was deactivated for a while, but the code wasn't removed. After reviewing this change thoroughly, I agree to remove the unused version management. It's a clever idea to use the Eclipse Do/Undo/Redo pattern.

PR #888, opened 2022-01-28
Fixed the molecule image being recreated for every draw cycle

### LINKED ISSUE #200: MoleculeImageService.create is a hot path when clicking peaks
image

Mailaender:
as this is expensive. Closes https://github.com/OpenChrom/openchrom/issues/200

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedMoleculeUI.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedMoleculeUI.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2021 Lablicate GmbH.
+ * Copyright (c) 2020, 2022 Lablicate GmbH.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -88,6 +88,7 @@ public class ExtendedMoleculeUI extends Composite implements IExtendedPartUI {
 	//
 	private double scaleFactor = SCALE_DEFAULT;
 	private Image imageMolecule = null;
+	private ILibraryInformation renderedLibraryInformation;
 	//
 	private IPreferenceStore preferenceStore = Activator.getDefault().getPreferenceStore();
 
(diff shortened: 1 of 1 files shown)


eselmeister:
The `imageMolecule = moleculeImageService.create(display, libraryInformation, width, height);` creates a new Image. Is there no need to dispose the existing image?

Mailaender:
Is 
https://github.com/eclipse/chemclipse/blob/3bc2434c948e91c32e0f6d067666784762a65aa0/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedMoleculeUI.java#L109-L111

not enough?

eselmeister:
I see a problem with the missing "dispose()" action here. System resources must be disposed, see rule 1:
[https://www.eclipse.org/articles/swt-design-2/swt-design-2.html](https://www.eclipse.org/articles/swt-design-2/swt-design-2.html)
_"If you created it, you dispose it."_

The molecule image service could cache the image, but that is IMHO a bit problematic as the image service has no knowledge, when to dispose the image. If the image is not disposed, we are running into a memory leak. If the image is cached and disposed by the callee, then it could happen that the image is still in use, which leads to a crash.

Hence, it's reasonable to follow the rule 1 and let `ExtendedMoleculeUI` decide when to dispose an image. I clearly see the point, that the image creation shall be cached somehow at this point. What do you think of the following solution:

Let the `ExtendedMoleculeUI` cache the images by using a hash map, stored by the library information hash code. If an image is requested, the map is asked for an existing image. If the map exceeds the size of 25 stored images, all mapped images are disposed and the map is cleaned before creating and adding the new image.

`private Map<Integer, Image> imageMap = new HashMap<>();`

As a class member.

```
moleculeInfo = getMoleculeInformation(libraryInformation);
if(isSourceDataAvailable(libraryInformation)) {
	
	int hashCode = libraryInformation.hashCode();
	imageMolecule = imageMap.get(hashCode);
	if(imageMolecule == null) {
		imageMolecule = moleculeImageService.create(display, libraryInformation, width, height);
		if(imageMap.size() > 25) {
			for(Image image : imageMap.values()) {
				image.dispose();
			}
			imageMap.clear();
		}
		imageMap.put(hashCode, imageMolecule);
	}
} else {
	logger.info(ERROR_MESSAGE);
}
```

Mailaender:
Sorry, I didn't know that replacing objects does not dispose them properly.

The hot path is not the user clicking on many peaks in quick succession, but the image getting generated multiple times whenever a new widget draw was triggered. I added a much simpler cache instead which silences this path:

image

eselmeister:
Two problems occur when using this change:

* It's not possible to select another molecule image service any longer
* The molecule info will be empty if the create molecule image operation is called twice

Why?

The molecule service is applied only once in the `initialize()` method. A selection of the service combo box has no effect. Currently CDK is the only provider, why this behavior is not visible. In the future, other services like PubChem might be added. Then it will fail. The other issue is the molecule info:

```
String moleculeInfo = "";
if(moleculeImageService != null) {
	...
	//
	if(renderedLibraryInformation == libraryInformation) {
		return;
	}
	// Info is empty ... but method has returned already
	moleculeInfo = getMoleculeInformation(libraryInformation);
	...
```

Mailaender:
Removed the caching of the molecule service. It may not be part of the hot path anyway.

PR #893, opened 2022-02-01
Increased the maximum MS database sorting limit

Mailaender:
@massbank is ~90.000 entries.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.msd.swt.ui/src/org/eclipse/chemclipse/msd/swt/ui/preferences/PreferenceSupplier.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.msd.swt.ui/src/org/eclipse/chemclipse/msd/swt/ui/preferences/PreferenceSupplier.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2018 Lablicate GmbH.
+ * Copyright (c) 2015, 2022 Lablicate GmbH.
  * 
  * All rights reserved.
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


eselmeister:
I recommend to not set this value beyond 30000 as JFace tables are getting increadible slow if too much data is loaded. 

```
private boolean isMassiveData(IMassSpectra massSpectra) {

	if(massSpectra != null) {
		int limitMassiveData = PreferenceSupplier.getLibraryMSDLimitSorting();
		return (massSpectra.size() > limitMassiveData);
	}
	return false;
}
```

JFace offers a virtual modus which is able to handle huge amounts of data. The drawback is, that sorting is disabled.

```
private void setLabelAndContentProviders(boolean massiveData) {

	setLabelProvider(labelProvider);
	if(massiveData && isVirtualTable()) {
		/*
		 * Lazy (Virtual)
		 */
		logger.info("Lazy (Virtual) Modus");
		setContentProvider(lazyContentProvider);
		setUseHashlookup(true);
		setComparator(null);
	} else {
		/*
		 * Normal
		 */
		logger.info("Normal Modus");
		setContentProvider(normalContentProvider);
		setUseHashlookup(false);
		setComparator(tableComparator);
	}
}
```

Setting this value to 1M will definitively cause problems when listing the contents of a MALDI-MS file in the Scan Table part.

Mailaender:
I am setting it to 100 k, though. What do you suggest instead?

eselmeister:
I'm not happy with a high value, cause even 30000 is problematic in regard to performance. Try to create a test scan with > 30000 m/z values and see how the "Scan Table" reacts. It will be definitely not "reactive".

Mailaender:
I tested it with @massbank. It was choppy, had to wait a few seconds for the sort to execute.

PR #894, opened 2022-02-02
Fixed directories getting scanned for file extensions

Mailaender:
If a directory ends with `.raw` it will trigger a lot of useless checks.


eselmeister:
That's problematic as vendors do store their binary file(s) also in directories, with a specific name, e.g.:

- Agilent *.D
- Waters *.raw

The customer is used to click on the vendor specific directory to open the chromatogram instead of clicking directly on the file. Moreover, some chromatograms are stored in one file, some in several files, some in a directory. It's a mess.

PR #896, opened 2022-02-02
Fixed AMDIS being displayed as a detector independent algorithm

### LINKED ISSUE #895: Processor Quick Access toolbar items are duplicated
e.g. Library File (MS) where one does not work as it is meant to be applied on mass spectra.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/895


eselmeister:
I do not agree to remove the "Scan Identifier" and "Scan Filter" operations from the quick access toolbar, as both are valid actions on the chromatogram. The scan identifier identifies the currently selected scan:

Bildschirmfoto von 2022-02-03 14-19-48

Selected ions are removed.

Bildschirmfoto von 2022-02-03 14-19-36

Mailaender:
Hmm, I see. I am not sure how to communicate this better through UI. Dropping https://github.com/eclipse/chemclipse/pull/896/commits/c26d3b6ac0804a97583f9125227849d9a596a96f completely then.

PR #897, opened 2022-02-03
Fixed *.cal editor not being marked as dirty when changes were made

Mailaender:
This is a followup of https://github.com/eclipse/chemclipse/pull/877


eselmeister:
In `EditorCalibration` please sort methods by access modifier public, protected, private. Please also take care to unsubscribe the event as soon as the editor is closed/finalized.

Mailaender:
Unsubscribed the event and added a dirty flag when renaming an entry. Sorted the members.

PR #903, opened 2022-02-15
Added undo/redo support for Denoising

Mailaender:
like https://github.com/eclipse/chemclipse/pull/892.


eselmeister:
There is still an issue. After applying the filter, the "undo" button is not active.

Mailaender:
You have to select/deselect the editor. It seems to be related to a larger UI update problem that I plan to address later.

eselmeister:
OK, that works. I'm fine if that will be fixed later on.

PR #907, opened 2022-02-17
Fixed performance of MALDI-TOF data import

Mailaender:
This removes a performance bottleneck where every ion was checked against all ions for every ion addition. It would take minutes to load a file with ~300.000 ions while other tools like http://www.mmass.org/ import them instantly.


eselmeister:
Principally, the change makes sense. But it leads to problems of failing test cases.

Bildschirmfoto vom 2022-02-17 12-57-04

Using `isTandemMS` ignores nominal m/z data. I can't recall exactly, why this code has been added and detailed test cases have been added. But my sixth sense says, this is important.

I think, that this can be solved easily. By default, the functionality is executed as is. The interface `IScanMSD` respectively `AbstractScanMSD` class offers a method to overwrite the option:

```
boolean ignoreIntensityCollisions(); // default false
```

The ignore coliision flag can be then set explicitly to true in the MALDI-TOF vendor scan.

What do you think?

Mailaender:
I agree that this may be converter/instrument or even vendor specific. `checkIon()` which is also a hot path seemed to check ion transitions so I thought that would be MS/MS specific.

PR #913, opened 2022-02-21
Fixed Peak/Scan list not updating when hidden in the background

### LINKED ISSUE #912: Peak/Scan List does not update on opening a new chromatogram
The content of the Peak/Scan List does not update when one chromatogram is currently active and a new one is loaded. 

tested cases (on OSX):
- active ch contains no peaks, new chromatogram contains peaks -> peaks new ch not shown until switch back/forth between the two ch
- active ch contains peaks, new chromatogram contains no peaks -> peaks from old ch still visible until switch back/forth between the two ch
- active ch contains peaks, new chromatogram contains peaks -> peaks from old ch still visible until switch back/forth between the two ch

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/912

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/parts/PeakScanListPart.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/parts/PeakScanListPart.java
@@ -56,6 +56,15 @@ public class PeakScanListPart extends AbstractPart<ExtendedPeakScanListUI> {
 		}
 	}
 
+	@Override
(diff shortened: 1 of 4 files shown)


eselmeister:
Looks good. I would additionally recommend to check with `instanceof` if the object is a chromatogram selection:

```
@Override
public boolean setFocus() {

	DataUpdateSupport dataUpdateSupport = Activator.getDefault().getDataUpdateSupport();
	List<Object> objects = dataUpdateSupport.getUpdates(IChemClipseEvents.TOPIC_CHROMATOGRAM_XXD_UPDATE_SELECTION);
	if(!objects.isEmpty() && objects.get(0) instanceof IChromatogramSelection) {
		chromatogramSelection = (IChromatogramSelection)objects.get(0);
	}
	updateChromatogramSelection();
	return true;
}
```

Mailaender:
Added the missing check.

PR #914, opened 2022-02-21
Added basic metadata for MALDI-TOF MS spectra

Mailaender:
This is a followup to https://github.com/eclipse/chemclipse/pull/911 which adds some metadata fields commonly found in data files.


eselmeister:
I would recommend to move the class `ExtendedMassSpectrumHeaderUI.java` to the package:

```
org.eclipse.chemclipse.ux.extension.xxd.ui.swt
```

Normally, scans of a chromatogram don't need variables like File or Header values. Hence, I would recommend to introduce a new interface on top of `IVendorMassSpectrum` probably name it `IVendorMassSpectrumTOF`

```
IVendorMassSpectrumTOF extends IVendorMassSpectrum
```

which contains the header values and the file member. Regarding the header values, have a look at the chromatogram. It implements `IMeasurementInfo`, which stores the header values as a map and thus is easily extensible. Specific header values like Date or Operator are pre-defined.

The TOF specific classes could look like this:

```
public class VendorMassSpectrumTOF extends AbstractVendorMassSpectrumTOF {...}
public abstract class AbstractVendorMassSpectrumTOF extends AbstractVendorMassSpectrum implements IVendorMassSpectrumTOF {...}
```

If TOF specific functionality is required, an `instanceof` could be used. Otherwise, the basic OpenChrom functionality can simply be used.

Mailaender:
I thought `IMassSpectrum` was already for single MS spectra. Indeed empty fields at this magnitude are an avoidable performance problem. I refactored it as suggested.

PR #920, opened 2022-02-23
Fixed Peak/Scan List not updating when adding scan identifiers

### LINKED ISSUE #918: Peak/Scan List not updated automatically from Scan Identifiers, Peak Detector
Currently, when a Scan Identifier or Peak Detector is run, while the Peak/Scan List is already open (independent whether it's currently visible or not) the Peak/Scan List is not updated. A manual press to the 'Reset Peak/Scan List' is required to visualise the Peaks/Scans in the List.

Requested behaviour would be that running an above operations will update the Peak/Scan List, ideally independent from whether it's currently visible or not. Having a GUI application, it feels counter intuitive that you have to press a reset button to obtain data. If there is no way to solve it, I suggest that 'update data' button would be a better name.

Tested on OSX, both on MSD and WSD chromatogram

Note that this is different from #913 which now works (opening/closing a chromatogram will fire an update on the Peak/Scan List).

Mailaender:
and scan identifiers not dirtying the chromatogram which makes this a followup to https://github.com/eclipse/chemclipse/pull/875.

Closes https://github.com/eclipse/chemclipse/issues/918

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.chromatogram.msd.identifier.ui/src/org/eclipse/chemclipse/chromatogram/msd/identifier/ui/core/ChromatogramFilter.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.chromatogram.msd.identifier.ui/src/org/eclipse/chemclipse/chromatogram/msd/identifier/ui/core/ChromatogramFilter.java
@@ -106,6 +106,7 @@ public class ChromatogramFilter extends AbstractChromatogramFilter {
 					public void run() {
 
 						MassSpectrumIdentifier.identify(massSpectra, massSpectrumIdentifierSupplier.getId(), monitor);
+						chromatogram.setDirty(true);
 					}
 				});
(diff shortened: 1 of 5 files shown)


eselmeister:
This change is problematic as it's not possible to select more than one peak in the peak list any longer. Additionally, the selection disappears immediately in the peak/scan list due to updating the chromatogram selection. The problem arises from this change:

Bildschirmfoto vom 2022-02-25 11-04-09

Please test it on your side. Try to select one or more peaks in the peak/scan list.

Mailaender:
That change is actually not at all required for this to work. I removed it.

eselmeister:
In the current change, two clicks are required to select a peak/scan in the list if the focus was lost. Is that intended?

Mailaender:
No, that is a recent regression. I seem to have introduced this in https://github.com/eclipse/chemclipse/pull/913. Added a fix.

PR #921, opened 2022-02-24
Cleaned up MassSpectrumIdentifier and SampleQuantProcessor

Mailaender:
This contains no real change.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.chromatogram.xxd.process.supplier.workflows/src/org/eclipse/chemclipse/chromatogram/xxd/process/supplier/workflows/core/SampleQuantProcessor.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.chromatogram.xxd.process.supplier.workflows/src/org/eclipse/chemclipse/chromatogram/xxd/process/supplier/workflows/core/SampleQuantProcessor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2021 Lablicate GmbH.
+ * Copyright (c) 2016, 2022 Lablicate GmbH.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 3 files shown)


eselmeister:
Adding the "`return false`" statements in the class `MassSpectrumIdentifier` changes the process logic. If such a case would have been intended, it would have been easier to simply write:

```
return supplier.isImportable();
```

But this would also change the process logic. As soon as the first supplier doesn't support the import, the check of all others is skipped as false is immediately returned. That's why only "true" is returned if the current supplier support the import. In all other cases, test each supplier until at least one supplier is available to support the import. The method return "`false`" at its end if no previous action returned true.

Mailaender:
Correct. Well spotted.

eselmeister:
The problem remains. It's the logic of the return statement. It could be, that the supplier matches, but that it is not importable. In such a case, false would be returned and all other supplier checks will be skipped.

Mailaender:
There is no logic change in there. See https://rules.sonarsource.com/java/RSPEC-1126 for the violation.

eselmeister:
For the pure statement `return supplier.isImportable();` this is correct. But it's the for loop that matters here:

```
for(ISupplier supplier : suppliers) {
	directoryExtension = supplier.getDirectoryExtension().toUpperCase();
	if(!directoryExtension.isEmpty() && directory.endsWith(directoryExtension)) {
		return supplier.isImportable();
	}
}
```
Imagine the `directoryExtension` is not empty and all supplier match the `directoryExtension`. Additionally, imagine we have three suppliers, but the first supplier doesn't have import capabilities:

- Supplier A - Importable: false
- Supplier B - Importable: true
- Supplier C - Importable: true

Then B and C will never be checked, as `supplier.isImportable();` returns `false` for the Supplier A.

Mailaender:
Updated.

eselmeister:
You are right. The logic was already wrong in the basic version. It seems, that something is odd with this class. I'll take care of it.

PR #923, opened 2022-02-24
Added alkane matching by CAS or IUPAC name

Mailaender:
This is much smarter for cases where there are already fully identified chromatograms.


eselmeister:
For a simple access of the `RetentionIndexCalculator`, I would recommend to keep the old method

```
public static int getAlkaneNumber(String name) {...}
```

and rename the new method e.g. to

```
public static int getAlkaneNumberEnhanced(ILibraryInformation libraryInformation) {...}
```

Mailaender:
Updated it to not break API.

PR #1046, opened 2022-05-23
Fixed toolbars icons being duplicated

Mailaender:
making them hard to differentiate.


eselmeister:
The proposed changes look good. But the integration icon is needed:
`String IMAGE_INTEGRATION_RESULTS = PATH_PREFIX + "integrationResults.gif";`

Even if integration and quantitation have the same icons at the moment, both are different process steps.

Mailaender:
I reused `IMAGE_CHROMATOGRAM_INTEGRATOR` instead.

eselmeister:
The `IMAGE_CHROMATOGRAM_INTEGRATOR` shall indicate, which kind of integrator is used.
chromatogramIntegrator

The `IMAGE_INTEGRATION_RESULTS` image is used to display that a certain result is available:
integrationResults

It would be great if we could keep the integration results image.

Mailaender:
Removed the change.

PR #1074, opened 2022-06-28
Fixed status line not displaying relevant information

Mailaender:
Replacing the redundant and broken perspective switch with feedback messages.


eselmeister:
I have tested the requested changes. The lower status line seems to be gone. Was this intended? Some issue occur regarding the lower borders now. Let's chat first so that you can show me the change in detail an give me further infos.

PR #1108, opened 2022-08-18
Improved performance around the Targets part

Mailaender:
I noticed that
> Targets have been modified (deleted/updated).

gets logged a lot and it triggers the `TOPIC_EDITOR_CHROMATOGRAM_UPDATE` which many parts listen to not just the editor, but also the Peak/Scan List which might then trigger another update.

Every selection in the Targets list would trigger it, even when no targets are actually chosen or renamed/deleted. This removes this update to reduce the load on the overall update mechanisms. It also does not seem to be required to update the automatic library search in the Scan Comparison part.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedTargetsUI.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedTargetsUI.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2022 Lablicate GmbH.
+ * Copyright (c) 2017, 2023 Lablicate GmbH.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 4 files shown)


chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedTargetsUI.java
```diff
@@ -786,8 +786,6 @@ private void updateOnFocus() {
 			Object object = objects.get(0);
 			if(object instanceof IChromatogramSelection chromatogramSelection) {
 				updateChromatogram(chromatogramSelection);
-			} else {
-				updateOther(object);
```
eselmeister, line 796: We should inspect this in more detail. The targets part displays targets (`IIdentificationTarget`) from various sources. The source implement the following interface `ITargetSupplier`:

- Chromatogram
- Peak
- Scan
- Editor

In this case, I assume that changes from an editor source are not fetched on focus.

Mailaender: https://github.com/eclipse/chemclipse/pull/1229 might be a better solution.

PR #1109, opened 2022-08-18
Improved Mass Spectra performance

Mailaender:
This is https://rules.sonarsource.com/java/RSPEC-1155 hidden behind custom interfaces that just seems to wrap a `List`. In the case of `IMassSpectra`, I assume it will be bad, as those databases can contain many entries. It should be even worse for `IIonProvider` because scans take up most of the RAM usage and item count when profiling OpenChrom.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.msd.identifier.supplier.nist/src/org/eclipse/chemclipse/msd/identifier/supplier/nist/core/support/Identifier.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.msd.identifier.supplier.nist/src/org/eclipse/chemclipse/msd/identifier/supplier/nist/core/support/Identifier.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2021 Lablicate GmbH.
+ * Copyright (c) 2008, 2022 Lablicate GmbH.
  *
  * All rights reserved.
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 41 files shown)


chemclipse/plugins/org.eclipse.chemclipse.msd.identifier.supplier.nist/src/org/eclipse/chemclipse/msd/identifier/supplier/nist/core/support/Identifier.java
```diff
@@ -140,9 +136,6 @@ public IMassSpectra runMassSpectrumIdentification(List<IScanMSD> massSpectrumLis
 					long maxProcessTime = (long)(searchSettings.getTimeoutInMinutes() * IChromatogramOverview.MINUTE_CORRELATION_FACTOR);
 					logger.info("Max Process Time: " + maxProcessTime);
 					logger.info("Run Identification");
-					Compounds compounds = runNistApplication(runtimeSupport, maxProcessTime, waitTime, monitor);
-					logger.info("Assign Compounds");
```
eselmeister, line 144: Is it intended to skip the identification process?

Mailaender: You are right. I removed too much. Only the `identificationResults` reassigned is unused.

PR #1130, opened 2022-09-15
Update Orbit to version 2022-06

Mailaender:
and remove unused dependencies.


eselmeister:
@Mailaender Are you sure that the dependencies can be removed? For example, `org.bouncycastle` ...  is required by `org.eclipse.chemclipse.pdfbox.extensions`.

Mailaender:
The import is not used in that plugin though.

Issue #1123, opened 2022-08-30
Calibration Editor - enable *.cal file selection with upper/lower case


eselmeister:
Currently, two entries for the same file extension are listed: one upper case, the other lower case. This should be merged, so that the user instantly can select the desired file, independent of the file extension upper/lower case.

eselmeister:
![Screenshot from 2022-08-30 15-32-45](https://user-images.githubusercontent.com/686671/187452125-25fff013-45a4-4952-ba60-227f2938ab98.png)

![Screenshot from 2022-08-30 15-33-13](https://user-images.githubusercontent.com/686671/187452134-6d6c3b41-7db8-44c7-b3e5-988584742a22.png)

PR #1134, opened 2022-09-15
Allow both upper and lower case for MS database file endings

Mailaender:
Essentially the same as https://github.com/eclipse/chemclipse/issues/1123 which are equally annoying.


eselmeister:
@Mailaender The license header need to be adjusted.

Issue #1186, opened 2022-11-14
Data Explorer - network attached drives on Microsoft Windows


eselmeister:
Often users complain that the application startup is slow, especially if many network attached drives are connected. We have seen, that this is severe if the network has bad response times and is heavily in use.

Hence, add an option to parse all network attached drives or only selected drives, like local ones. By default, only local drives shall be listed. The user must actively deactivate "List Drives by Type" to show all drives. Network drives can be also accessed via the "User Location".

This should speed up the startup time.

eselmeister:
![Screenshot from 2022-11-14 07-11-00](https://user-images.githubusercontent.com/686671/201588988-37d31f70-a1a0-4ec3-b551-63137f44ab16.png)

![Screenshot from 2022-11-14 07-11-22](https://user-images.githubusercontent.com/686671/201588992-acda716b-cf90-4b97-9b62-f64eb0f61be9.png)

PR #1187, opened 2022-11-14
Fixed no Windows drives but C:\ being shown in the Data Explorer

Mailaender:
This reverts https://github.com/eclipse/chemclipse/issues/1186 because it didn't work. The file system type is not local vs shared on Windows 10, but rather NTFS vs. FAT32 so everything was falsely detected as a network drive leaving the file explorer empty but the back fall to C:\ which might not even be allowed to access in restricted environments.

image

image

Removing default access to network drives is not an option. Usually, this is the main data source for laboratories. Hiding that behind multiple complicated options is not optimal. I therefore always assume that network drives are not empty and only check for children in the network drive root folder when the user explicitly expands the tree viewer hoping this improves application startup performance.


eselmeister:
@Mailaender You're right, it doesn't detect a local or remote (attached network drive) in a reliable way. The documenation already warned, that the type information is highly dependent on the implementation:

```
FileStore fileStore = Files.getFileStore(root.toPath());
String type = fileStore.type();
```
See:

```
/**
* Returns the <em>type</em> of this file store. The format of the string
* returned by this method is highly implementation specific. It may
* indicate, for example, the format used or if the file store is local
* or remote.
*
* @return  a string representing the type of this file store
*/
public abstract String type();
```
But I disagree to just revert the change as it not solves the problem that user reported long startup times. IMHO, it only happens on Windows, as the drives are attached on root level. On Linux and macOS, attached network drives are usually under `/mnt` or `/media` and not evaluated when setting up the "Data Explorer".

I would be fine to extend the extend the preference text field `P_WINDOWS_DRIVE_TYPE` and let the user type in a list of drives (local/remote) that shall be displayed. The default value for the text field could be to list:

- C:
- D:
- E:

If the user wants to show all drives on Windows, the option to deactivate listing the drives by name is still available. Furthermore, via "User Location" the user can freely select any attached network drive.

Let's me prepare a solution an push it today.

I would be happy if we find a way to clearly determine on Windows whether a drive is a local or remove one, but I didn't find a solution to achieve this yet.

Mailaender:
Updated so this becomes a toggleable setting again.

PR #1211, opened 2022-12-12
Renamed hidden home folder directory name

Mailaender:
The current setup creates `~/.chromatographyapplication/` which is pretty odd for several reasons: it is not the application name, this is not just about chromatography and the `_` had to be removed.


eselmeister:
Via the launching arguments in the `.product` file, the use can decide where to store the application data:

```
-Dosgi.instance.area=@user.home/.myapp/0.9.x
-Dosgi.user.area=@user.home/.myapp/0.9.x
```

Hence, this change should be fine.

PR #1248, opened 2023-01-18
Fixed unneccessary casts failing the build

Mailaender:
```
[ERROR] chemclipse/plugins/org.eclipse.chemclipse.xxd.process.supplier.pca.ui/src/org/eclipse/chemclipse/xxd/process/supplier/pca/ui/internal/wizards/PeakSettingsWizardPage.java:[210] 
[ERROR] 	if(o1 instanceof Integer i) {
[ERROR] 	   ^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Expression type cannot be a subtype of the Pattern type


eselmeister:
Ah yep, you're right.

PR #1249, opened 2023-01-18
Documented zoom hotkeys

Mailaender:
probably inherited from SWT Chart.


eselmeister:
It's just an improvement in the HTML file. The CI fails, but it is related to another issue, which is already resolved.

PR #1251, opened 2023-01-24
Added a Pseudo Gel view

Mailaender:
which is basically a heatmap with stacked mass spectra painted in black and white.


Mailaender:
https://www.researchgate.net/project/OpenTyper/update/63d114d997e2867d507c2242

PR #1257, opened 2023-01-26
Added Drag & Drop to the Chromatogram Batch Editor

### LINKED ISSUE #161: Drag&Drop Support for Batch-Editor
It would be nice if files could be dragged and dropped from Data Explorer into the Batch Processor.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/161


eselmeister:
The wizard automatically validates, that the selected files contain a valid chromatogram. Using D&D is also fine and we should merge it. But it could lead to a situation, that a certain file can't be processed. But that's ok, as each file is handled separately and if it fails, it shouldn't affect the processing of the other chromatograms.

PR #1290, opened 2023-02-14
Multi Molecule Image Service

Mailaender:
This fixes a collection of issues arising for having more than one image service since https://github.com/eclipse/chemclipse/pull/1286 and https://github.com/eclipse/chemclipse/pull/1288.


eselmeister:
This change looks good, but it leads to an error in the OpenChrom/CDK package:
`net.openchrom.xxd.identifier.supplier.cdk.ui.services`

```
Description	Resource	Path	Location	Type
The type MoleculeImageService must implement the inherited abstract method IMoleculeImageService.isOnline()	MoleculeImageService.java	/net.openchrom.xxd.identifier.supplier.cdk.ui/src/net/openchrom/xxd/identifier/supplier/cdk/ui/services	line 36	Java Problem
```
Please create a PR in the OpenChrom/CDK package, so that both PRs can be merged in sync. Additionally, a public abstract class could be added, which defines "isOnline" by default as false. Any implementation, that uses online services, shall override to true.

`public abstract class AbstractMoleculeImageService implements IMoleculeImageService`

Mailaender:
I can push to @OpenChrom without your help, so this shouldn't be necessary. Creating an `AbstractMoleculeImageService` only to default it to something that may be wrong and has to be overwritten sounds like bad design.

PR #1304, opened 2023-02-21
Localized org.eclipse.chemclipse.chromatogram.csd.*

Mailaender:
https://github.com/OpenChrom/openchrom/issues/24


eselmeister:
The "empty" plugins are used as upper level preference page containers via "category":

Screenshot from 2023-02-22 06-17-07

Screenshot from 2023-02-22 06-17-28

PR #1397, opened 2023-05-22
Update to EJML 0.43

Mailaender:
https://github.com/lessthanoptimal/ejml/blob/v0.43/change.txt


lorenzgerber:
@Mailaender Tested with all unit tests and manual runs. No problems could be found.

Mailaender:
Note: I did this update mainly because it cleared the INFO level warnings for me.

PR #1409, opened 2023-06-07
Reduced rawtypes and unchecked warnings

Mailaender:
This is a followup to https://github.com/eclipse/chemclipse/pull/1408 which still does not solve the whole problem, but reduces the warnings further.


eselmeister (commented):
OK

eselmeister (approved):
Looks good.

PR #1418, opened 2023-06-14
Added more decimal places to exact masses from mzXML

Mailaender:
and adapt the UI to handle it.


chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ScanChartUI.java
```diff
@@ -674,8 +674,7 @@ private String getLabel(double value) {
 				}
 				break;
 			case EXACT:
-				DecimalFormat decimalFormat = getDecimalFormatValue();
```
eselmeister, line 677: Some vendor import might read the high res mass as is, so using Double.toString(value) could lead to long labels here. I would recommend to constrain it to e.g. 6 decimals. The decimal format allows to define a pattern, e.g. "0.000###". Show 3 decimal places in every case and up to 6 if available.

Mailaender: I already found a hardcoded limit of 6 decimals at https://github.com/eclipse/chemclipse/blob/0728b0e0776722179a2768539d3f42ceef86d099/chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/core/AbstractIon.java#L38

PR #1428, opened 2023-06-20
Reduced memory footprint of HPLC-DAD chromatograms

Mailaender:
Wavelengths are typically stored with just one decimal place, so double precision is over the top.


eselmeister (approved):
OK

PR #1431, opened 2023-06-22
Improved File Explorer performance by doing deep content checks only on click

### LINKED ISSUE #1412: File content matcher on click
The magic code checks are currently also used to parse the file and decide which type of content it contains. Maybe we should not do that for every file in the directory, but only the highlighted one to improve Data Explorer performance.

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/1412


eselmeister (approved):
Smart idea.

Issue #1447, opened 2023-07-07
Chromatogram Editor - replace JFace Actions/Toolbar


eselmeister:
The current chromatogram editor toolbar is hard to maintain and doesn't follow the ChemClipse toolbar style. Hence, replace it by keeping the functionality.

eselmeister:
Furthermore, the chromatogram retention indices list shall be also extracted into a separate part.

![Screenshot from 2023-07-07 09-11-17](https://github.com/eclipse/chemclipse/assets/686671/fe0508b6-8432-4cb6-b740-92c32c34679f)
![Screenshot from 2023-07-07 09-11-44](https://github.com/eclipse/chemclipse/assets/686671/d4955e83-aa4b-4224-a9fa-c359a369d442)
![Screenshot from 2023-07-07 09-12-15](https://github.com/eclipse/chemclipse/assets/686671/6a660127-b690-4be7-8476-da8a24588405)

eselmeister:
Furthermore, the chromatogram retention indices list shall be also extracted into a separate part.

eselmeister:
![Screenshot from 2023-07-10 13-45-37](https://github.com/eclipse/chemclipse/assets/686671/8b992737-6c63-442a-b507-9e5cd77b45fc)
![Screenshot from 2023-07-10 13-46-20](https://github.com/eclipse/chemclipse/assets/686671/264d4cfe-46d6-4854-aa3e-9b77b67cc12c)
![Screenshot from 2023-07-10 13-46-37](https://github.com/eclipse/chemclipse/assets/686671/d976197f-8eea-4327-9c19-81bbd6a35751)
![Screenshot from 2023-07-10 13-47-00](https://github.com/eclipse/chemclipse/assets/686671/8050588c-7910-43af-b957-8c5f5a329c58)

PR #1452, opened 2023-07-11
Added native toggle buttons

Mailaender:
https://github.com/eclipse/chemclipse/issues/1447 inspired me to rework the icons. Instead of overlaying a :red_circle: I simply used [SWT toggle buttons](https://www.eclipse.org/swt/widgets/). This also allows tracking the state on the button and feels less alien to the user. The `MDirectToolItem` will ignore the check state when it also gets a menu, so I repainted the icons with a dark inactive state. Not having 25% of them covered with a red dot on a red icon makes them a bit easier to distinguish.


eselmeister:
That looks principally good. But the icons in the main toolbar are grey, so people might think they are disabled.

Screenshot from 2023-07-11 16-08-51

Mailaender:
What do you suggest instead?

eselmeister:
Is it possible to keep the color of the icon, so that it behaves like the breakpoint toggle in Eclipse?

Screenshot from 2023-07-11 22-02-31

Screenshot from 2023-07-11 22-02-36

When the category is active, the gray background shall be displayed.

Mailaender:
> The `MDirectToolItem` will ignore the check state when it also gets a menu...

Mailaender:
I tried hacking around this using a checked button and fake dropdown next to it in a tool control, which, however, required a lot of custom code and gave very flaky results. I now pasted the Eclipse pin needle into the icons, similar to the legacy ones.

eselmeister:
The changes of this PR have been cherry-picked and pushed via:
https://github.com/eclipse/chemclipse/issues/1460

PR #1522, opened 2023-10-04
Added support for settings with 8 or 16 bit data types

Mailaender:
including a test case where a value between 1 and 100 is now stored as byte.


eselmeister:
@Mailaender Using short and byte as data types looks good. When changing

```
@IntSettingsProperty
private int numberOfTargets = PreferenceSupplier.DEF_NUMBER_OF_TARGETS;
```

to

```
@ByteSettingsProperty(minValue = PreferenceSupplier.MIN_NUMBER_OF_TARGETS, maxValue = PreferenceSupplier.MAX_NUMBER_OF_TARGETS)
private byte numberOfTargets = PreferenceSupplier.DEF_NUMBER_OF_TARGETS;
```

Can you ensure backward compatibility?
The setting is already used, so we don't want to break existing/used settings.

Mailaender:
Internally, it is still saved as int because the Eclipse preferences framework does not support short. JSON only supports a generic number data type. This is more about validating min/max ranges properly when the Java data type is smaller.

PR #1525, opened 2023-10-06
Fixed optional setting fields requiring a value set

Mailaender:
https://github.com/eclipse/chemclipse/pull/1523 introduced a regression where everything was required because I reused a validator that is put on every widget.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.support/src/org/eclipse/chemclipse/support/settings/validation/EmptyStringValidator.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.support/src/org/eclipse/chemclipse/support/settings/validation/EmptyStringValidator.java
@@ -1,0 +1,37 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Lablicate GmbH.
+ * 
+ * All rights reserved.
+ * This program and the accompanying materials are made available under the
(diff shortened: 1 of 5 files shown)


chemclipse/plugins/org.eclipse.chemclipse.support/src/org/eclipse/chemclipse/support/settings/validation/EmptyStringValidator.java
```diff
@@ -0,0 +1,34 @@
...
+	private static final String ERROR = "Please enter a value.";
+
+	public EmptyStringValidator() {
+
+	}
+
+	@Override
+	public IStatus validate(Object value) {
+
+		if(value instanceof String text && text.isEmpty()) {
```
eselmeister, line 32: A `if(value == null) {}` is required here too.

eselmeister:
The `EmptyStringValidator` requires a null check:

Currently, ValidationStatus.ok() is returned if value is null.
```
@Override
public IStatus validate(Object value) {

	if(value instanceof String text && text.isEmpty()) {
		return ValidationStatus.error(ERROR);
	}
	return ValidationStatus.ok();
}
```

Mailaender:
That is already checked on `InputValidator`, but I can add it here, too.

Issue #1666, opened 2024-02-28
Header Data - add findings support


eselmeister:
Add a fixed header data field to store findings and display the data in the header data part.

eselmeister:
![Screenshot from 2024-02-28 17-48-59](https://github.com/eclipse/chemclipse/assets/686671/fe6d50d3-dbe5-43a2-b996-d03008cb01e4)

PR #1667, opened 2024-02-28
#1666 header data

eselmeister:


Mailaender:
Seems very much redundant to Misc Info, Misc Info Separated, Short Info, Detailed Info...

eselmeister:
@Mailaender You're right. Within the next months, we should unify the header data table.

PR #1676, opened 2024-03-02
Remove German words in English translation

Mailaender:
https://dict.leo.org/englisch-deutsch/extern


eselmeister:
@Mailaender The `NIST_IDENTIFIER` is used as an identifier to load the reference mass spectrum from a database.  It is stored in the target and saved. If it is translated, then the library service does not know, which reference database to load.

PR #1685, opened 2024-03-08
Remove OpenChrom trademark

Mailaender:
like https://github.com/eclipse/chemclipse/pull/860 did.


eselmeister:
Isn't a correct automatic module name required?
```
org.eclipse.chemclipse.xxd.identifier.supplier.wikidata.ui
...
```

Mailaender:
Apparently not when it was the wrong one all the time.

eselmeister:
Probably, it's better to set the correct one instead of none to be consistent with all other bundles.

Mailaender:
Fixed it in a few other occasions.

PR #1730, opened 2024-04-18
Remove LTK and PDE/JDT JUnit runtime

Mailaender:
I can run unit tests without them and [LTK](https://wiki.eclipse.org/FAQ_What_is_LTK%3F) is unused IDE functionality. I am not sure why I added it 2 years ago https://github.com/eclipse/chemclipse/pull/1144 but those are not required anymore. JDT is also not needed for the HTML help anymore. https://github.com/eclipse/chemclipse/pull/1215


eselmeister:
As far as I remember, we played with the option to execute Java code from within "Project Explorer".

Issue #1796, opened 2024-07-02
Header part shows no data and crashes


Mailaender:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.nebula.widgets.richtext.RichTextEditor.setText(String)" because the return value of "java.util.concurrent.atomic.AtomicReference.get()" is null
	at org.eclipse.chemclipse.ux.extension.xxd.ui.swt.ExtendedHeaderDataUI.updateData(ExtendedHeaderDataUI.java:391)
	at org.eclipse.chemclipse.ux.extension.xxd.ui.swt.ExtendedHeaderDataUI.updateInput(ExtendedHeaderDataUI.java:364)
	at org.eclipse.chemclipse.ux.extension.xxd.ui.swt.ExtendedHeaderDataUI$4.widgetSelected(ExtendedHeaderDataUI.java:202)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:265)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5855)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1617)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:5065)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4517)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:152)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:639)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:546)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:173)
	at org.eclipse.chemclipse.rcp.app.ui.internal.support.ApplicationSupportDefault.start(ApplicationSupportDefault.java:25)
[...]

PR #1797, opened 2024-07-03
WiP #1796 - Header part shows no data and crashes

eselmeister:
An option has been added to use the rich or plain text editor. Additionally, a check could be added to test if WebKit is available on the system.


Mailaender:
Instead of yet another checkbox, it might be better to just display the error message of the `SWTError` in the control or in the feedback tab. Otherwise, you have to default to plain text editor to avoid the crash.

eselmeister:
By default, a plain text editor displays the data now. If the user likes to activate the RTF editor, then on demand. It's a way to ensure, that the header data is displayed on platforms where WebKit is not available.

PR #1846, opened 2024-08-31
Remove accidental XML m/z rounding

### LINKED ISSUE #1848: .mzML files: m/z values are imported as nominal masses (integer) and not required exact masses (numeric)
When opening _.mzML_ files (in _OpenChrom_), any digits following the decimal separator currently seem to be dropped from _m/z_ values. This effectively converts exact masses (numeric / float) to nominal masses (integers).

This is problematic for at least three reasons:
1. The nominal mass of larger analyte ions of a charge of |_z_| > 1 can not be approximated simply by taking the integer of their _m/z_ value. For example, consider the [_M_+2H]^(2+) ion (_z_ = 2) of the peptide LVRPWTSK (_M_ = 985.5709 Da), which has an _m/z_ of 493.79325 (rounded: 494), and not 493. When inferring the analyte's mass _M_ (Da), this results in an artificial mass defect that grows with the absolute value of the ion charge.
2. The charge |_z_| of unknown analyte ions is inferred from their isotope distribution; the distance (_m/z_) between its isotope peaks is the approximate reciprocal of their charge. For example, a distance of ~0.34 _m/z_ would indicate an analyte ion charge of |3|. Knowledge of the absolute value of the charge is required in order to calculate the analyte's mass _M_ (Da).
3. The unambiguous identification of bio-oligomers, such as glykans and peptides, requires high mass accuracy in order to differentiate between similar but distinct fragment ions. For example, consider the minute differences between the monoisotopic residue masses of following di-peptides: MM (262.0810 Da), DF (262.0954 Da), and YV (262.1317 Da).

In generic ESI-MS/MS analyses, ToF or orbitrap mass analyzers typically record _m/z_ values with an accuracy of 4 to 6 digits following the decimal separator.
Thus, I suggest to let _ChemClipse_ import _m/z_ values with an accuracy of at least 4 digits post separator. Missing digits could be replaced by zeros.

Mailaender:
as it leads to nominalized raw data. Closes #1848


LouisPM:
Thank you for having already submitted a fix, and let me apologize for my now duplicate report of issue #1848.

Mailaender:
No worries.

PR #1854, opened 2024-09-04
Improved Selection of OPLS Y

lorenzgerber:


eselmeister:
The change looks reasonable. I request two changes:

- Use `AtomicReference<ComboViewer>` to have the same structure as the other UI elements
- Make the `ComboBox` a bit larger by default 

Screenshot from 2024-09-04 22-07-30


```
GridData gridData = new GridData();
gridData.widthHint = 250;
combo.setLayoutData(gridData); 
```

PR #1868, opened 2024-09-12
Add chromatography type to chromatogram

Mailaender:
We currently differentiate by detector type, but not whether the chromatogram is from an LC or GC. Not all file formats are agnostic in this regard. Also, some types might want their own editor. This is a proposal. I am not sure if this is the right way to do it.


Mailaender:
I still think this is handy, but I am retracting it because I don't know against what to map this. For example mzXML had [separationTechnique](http://tools.proteomecenter.org/formats/mzXML/mzXML_xmlspy_docs.html#element_separationTechnique_Link02065950) but it is not defined by the standard and the external schema also leads to nowhere. https://web.archive.org/web/20040905052014/http://sashimi.sourceforge.net/schema_revision/mzXML_2.1/separations/separation_technique_1.0.xsd

PR #1549, opened 2023-11-13
Some minor updates to the Explorer UI

Mailaender:
including localization and avoiding redundancy in the labels to save some horizontal space.

PR #1932, opened 2024-11-05
Show the Peak List by default

Mailaender:
but focus the Data Explorer because people need to open their files first. The rationale is that the Peak List is far too important to hide in the most crowded toolbar menu. Since https://github.com/eclipse/chemclipse/pull/1549 we have more vertical space to allow for a fourth tab.


eselmeister:
In a first try, I stumbled and wondered why I'm not able to activate the peaks list part even though its header is visible. In a second look I saw, the the focus is set on the data explorer. I'm not sure, if this more helps or confuses the customers. I was also confused.

Mailaender:
When I keep its default focus, people will see an empty peak list when opening the Data Analysis view for the first time and then complain that they don't know how to open files or use the awkward ways from the Files menu instead.

eselmeister:
The *.e4xmi "Position in list:" option allows to place items via keywords before or after a specific part, e.g.:
`after:org.eclipse.chemclipse.ux.extension.xxd.ui.part.dataexplorer`

image

But I also need to have a look if that is possible here, as the part is created dynamically.

Mailaender:
Looks like this for me:
grafik
Changing `Selected Element` had no effect.

PR #1960, opened 2024-11-17
Use the default UI font on the welcome screen

Mailaender:
instead of Arial which is a proprietary font by @microsoft.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.ui/src/org/eclipse/chemclipse/ux/extension/ui/swt/TaskTile.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.ui/src/org/eclipse/chemclipse/ux/extension/ui/swt/TaskTile.java
@@ -80,12 +80,6 @@ public class TaskTile extends Composite {
 		updateFromDefinition();
 	}
 
-	@Override
-	public void dispose() {
-
-		super.dispose();
-	}
-
 	public TileDefinition getDefinition() {
 
 		return definition;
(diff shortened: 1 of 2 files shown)


chemclipse/plugins/org.eclipse.chemclipse.ux.extension.ui/src/org/eclipse/chemclipse/ux/extension/ui/swt/TaskTile.java
```diff
@@ -254,16 +255,21 @@ private void updateStyle(int style) {
...
-			fontSize = 18;
 		}
-		//
-		textSection.setFont(Fonts.getCachedFont(getDisplay(), "Arial", fontSize, SWT.BOLD)); //$NON-NLS-1$
+		FontData[] fontDatas = label.getFont().getFontData();
+		for(FontData fontData : fontDatas) {
+			fontData.setHeight(fontSize);
+			fontData.setStyle(SWT.BOLD);
+		}
+		textSection.setFont(new Font(getDisplay(), fontDatas));
```
eselmeister: Does the font need to be disposed?
https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fswt%2Fgraphics%2FFont.html

Mailaender: I think you are right however we never dispose the welcome view.

eselmeister:
Why do you create the `FontRegistry`, instead of using the existing approach via `Fonts`?
E.g. to avoid the Windows font name?
`Fonts.getCachedFont(getDisplay(), label.getFont().getFontData()[0].getName(), fontSize, SWT.BOLD)`

Mailaender:
I don't think that the font size is already cached and I don't see where that is disposing fonts.

eselmeister:
`Fonts` takes care of caching and disposing if needed. Do you dispose it via `FontRegistry`?

Mailaender:
https://github.com/eclipse/chemclipse/blob/3eaa42707b827c81e1081782cb457a6248929b0d/chemclipse/plugins/org.eclipse.chemclipse.swt.ui/src/org/eclipse/chemclipse/swt/ui/support/Fonts.java#L26 are never disposed.

[FontRegistry](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fjface%2Fresource%2FFontRegistry.html) will automatically dispose when the display is disposed so we don't have to.

eselmeister:
I don't like to introduce two concepts with this approach. I would then recommend to use the `FontRegistry` inside of `Fonts`, so that developer can easily rely on the Fonts class instead of creating a new `FontRegistry` whenever they get in touch with using fonts.

PR #1967, opened 2024-11-22
PCA, score plot, fix closest distance calculation

lorenzgerber:
implement euclidean distance based closest point algo for score plot


chemclipse/plugins/org.eclipse.chemclipse.xxd.process.supplier.pca.ui/src/org/eclipse/chemclipse/xxd/process/supplier/pca/ui/swt/ExtendedLoadingsPlot.java
```diff
@@ -467,9 +467,9 @@ private void applySettings() {
 		/*
 		 * Prevent Zoom reset on DblClick
 		 */
-		if(dblClicked){
-			plotControl.get().updateRange(rangeX, rangeY);	
-		} 
+		if(dblClicked) {
```
eselmeister: `dblClicked` -> try to avoid abbreviations, it's harder to read. Use instead, e.g.: `doubleClick`

PR #1972, opened 2024-11-24
Fixed Dark Mode detection on macOS

### LINKED ISSUE #1971: macOS automatic dark mode has unreadable font colors in charts
image

Mailaender:
Closes https://github.com/eclipse/chemclipse/issues/1971


chemclipse/plugins/org.eclipse.chemclipse.support.ui/src/org/eclipse/chemclipse/support/ui/workbench/PreferencesSupport.java
```diff
@@ -36,7 +41,13 @@ public static void cleanPreferencesByNodeId(List<String> preservePreferenceNodes
 	// Workaround for https://bugs.eclipse.org/bugs/show_bug.cgi?id=564022
 	public static boolean isDarkTheme() {
```
mickaelistria, line 43: I guess you mean "isDarkSystemTheme" here? If so, then I believe the new strategy should be the default one for most OSs, and Windows would be the exception as it basically doesn't set system colors.

PR #2022, opened 2025-01-14
Add detector type icons in editor tabs

Mailaender:
to get some vertical space back since the column type is also suffixed now:
grafik
and regain some consistency with the 3.X editor from projects that does not add these suffixes.


eselmeister:
Adding distinct icons for different detector types is a good idea. But to display the detector type `[CSD], [MSD], [VSD], [WSD]` helps to keep the overview if a chromatogram with the same name but different detectors is open in the editor tab. Hence, I would like to keep it.

PR #2037, opened 2025-01-17
Remove generics from PeakIdentifier

Mailaender:
My point is that the return type is always the same so there is no real need for generics. Removing them again makes the API clearer and avoids suppressed warnings.


eselmeister:
It's bad practice to remove contributors from the header. Please keep them even if functionality is modified.

eselmeister:
Additionally, add your name in the header when you make contributions.

Mailaender:
Functionality is not modified. The change was reverted, so the line does not make sense anymore. It is still visible in the commit history, so nothing was deleted. That is why I don't think these lines are of any use. I reverted their removals in this pull request for the sake of conventions.

PR #2049, opened 2025-01-24
Resort Peak/Scan List columns

Mailaender:
basically putting `Area %` and `Best Target` at first glance because they answer the question "what is it?" and "how much is it?" in an easy to parse manner. They were buried too deeply, imho. Also resized the columns so there is as much space as possible for the substance name:

grafik


eselmeister:
The `getScanText(IScan scan, int columnIndex) {...}` needs to be also adjusted.

Screenshot From 2025-01-27 20-00-38

Mailaender:
Yes, of course. Well spotted.

PR #2075, opened 2025-02-04
Feature: Toggle 'Use' tickbox, activate all samples

lorenzgerber:
Right click in the PCA Sample table to toggle 'use' for selected samples. Also, right click to reactivate all samples for analysis.


eselmeister:
Looks good. To write the code more compact, you could use:
`sample.setSelected(!sample.isSelected());`

PR #2082, opened 2025-02-10
Improve peak matching in Target Transfer

Mailaender:
This reduces some edge cases while transferring peaks. Instead of choosing the first retention time overlap, I choose the nearest peak within retention overlap to improve targeting. To avoid mislabelling, I also copy the original match factor to get the target sorting right. You can still override the match factor with a fixed value. Also fixes the manual verification tick not being transferred, which also broke the peak labelling. While not perfect, this gives much better results when transferring deconvoluted MS peaks onto a FID with first derivative (CB) peak detection.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.xxd.filter/src/org/eclipse/chemclipse/xxd/filter/chromatogram/settings/TargetTransferFilterSettings.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.xxd.filter/src/org/eclipse/chemclipse/xxd/filter/chromatogram/settings/TargetTransferFilterSettings.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024 Lablicate GmbH.
+ * Copyright (c) 2024, 2025 Lablicate GmbH.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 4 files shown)


chemclipse/plugins/org.eclipse.chemclipse.xxd.filter/src/org/eclipse/chemclipse/xxd/filter/chromatogram/settings/TargetTransferFilterSettings.java
```diff
@@ -27,10 +27,10 @@ public class TargetTransferFilterSettings {
 	@JsonProperty(value = "Transfer Best Target Only", defaultValue = "false")
 	@JsonPropertyDescription(value = "If this value is true, only the best target will be transfered.")
 	private boolean useBestTargetOnly = false;
-	@JsonProperty(value = "Match Quality", defaultValue = "80.0")
-	@JsonPropertyDescription(value = "The match quality is set as the Match Factor.")
+	@JsonProperty(value = "Match Quality", defaultValue = "")
```
eselmeister, line 30: Why don't you set a default value?

Mailaender: I want it to default to 0 (or leaving the field empty) so the match quality is taken from the source which gives correct ordering.

PR #2089, opened 2025-02-17
Sync tycho-build with main tycho version

akurtakov:
Should reduce warnings/errors printed on build start like:
```
[WARNING] The POM for org.apache.commons:commons-lang3:jar:3.10 is
invalid, transitive dependencies (if any) will not be available: 1
problem was encountered while building the effective model for
org.apache.commons:commons-lang3:3.10
[FATAL] Non-parseable POM
/home/jenkins/agent/workspace/chemclipse_develop/.mvn/org/apache/commons/commons-parent/50/commons-parent-50.pom:
UTF-8 BOM plus xml decl of ISO-8859-1 is incompatible (position:
START_DOCUMENT seen <?xml version="1.0" encoding="ISO-8859-1"... @1:42)
@ /home/jenkins/agent/workspace/chemclipse_develop/.mvn/org/apache/commons/commons-parent/50/commons-parent-50.pom,
line 1, column 42
```

Reference to the old tycho brings old plexus-util as transitive dependency and it ends up with problem https://github.com/codehaus-plexus/plexus-utils/commit/bb346ad63efc6f0cbfd5b1f1736df677bad276cf .

### DIFF
--- a/chemclipse/.mvn/extensions.xml
+++ b/chemclipse/.mvn/extensions.xml
@@ -3,6 +3,6 @@
   <extension>
     <groupId>org.eclipse.tycho</groupId>
     <artifactId>tycho-build</artifactId>
-    <version>2.7.0</version>
+    <version>4.0.11</version>
   </extension>
 </extensions>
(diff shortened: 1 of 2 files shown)


akurtakov:
I see a number of these failures in https://ci.eclipse.org/chemclipse/job/chemclipse/job/develop/3934/consoleFull

PR #2120, opened 2025-03-01
Add target view to MALDI-TOF MS perspective

Mailaender:
grafik


eselmeister:
Instead of Copy & Paste source code, it probably better should be unified.

Mailaender:
Yes, but it wasn't abstracted enough for that and had too many dependencies. I intend to do a chemical vs. biological targets and then revisit this crowded UI part.

PR #2121, opened 2025-03-02
Add a settings dialog to MALDI identifiers

Mailaender:
same as https://github.com/eclipse-chemclipse/chemclipse/pull/2111 and remove file format extensions restrictions from the file identifier.


eselmeister:
Removing the available file extensions to import spectra might lead to selecting wrong files, which could crash the identifier.

Mailaender:
Yes, ideally this would be auto-populated with available file extensions, however the Jackson annotation is not flexible enough.

PR #2127, opened 2025-03-05
Remove unused adapter interface from chromatogram and ion

Mailaender:
Careful putting too much stuff onto ion as it is a very frequently occurring object.


Mailaender:
Might be loosely coupled to something.

PR #2147, opened 2025-03-25
PCA, long format text import

lorenzgerber:


eselmeister:
@lorenzgerber I would rather suggest to name the field `sampleDetails` instead of `sampleNameLong`, which implies that it is only related to the name.

PR #2164, opened 2025-04-10
Remove PARAFAC peak import

Mailaender:
as it relies on an interaction with an undocumented @matlab script that is over 10 years old now.


eselmeister:
It's related to OpenFluor. As long as it does no harm, I would like to keep it.

PR #2169, opened 2025-04-23
Fix inconsistent duplicated target highlighter in Peak/Scan list

Mailaender:
In https://github.com/eclipse-chemclipse/chemclipse/pull/2153 I thought using the CAS to flag duplicates is a good idea like we do in peak identifications. However, this also marks substances across split peaks that are actually correctly assigned.

This also fixes scans not being flagged as duplicates and best target in Peak/Scan list staying empty for them.


eselmeister:
The ECA check failed as the Eclipse services show a massive outage currently.
image

Manually confirm that @Mailaender is a valid comitter and has signed the ECA. Nothing has changed since the last commits.

PR #2178, opened 2025-05-06
Remove .0 suffix for nominal m/z in Scan Chart

Mailaender:
to avoid some visual noise as the ticks are labelled with integer like ranges.

Before:
grafik

After:
grafik


eselmeister:
It looks good for nominal m/z data, but leads to odd results when showing high resolution m/z values.

Screenshot From 2025-05-07 08-53-59

Did you consider to add settings options for the different types to let the user specify the format for each type NOMINAL, TANDEM, HIGH_RES in the settings like it is done for the chromatogram axes:

Screenshot From 2025-05-07 08-54-29

Mailaender:
Added the decimal place back for high-res and tandem MS. I think this should be set automatically based on data and instead of making the user correct it for every file opened like we do for the label on top of the peaks. Also, these decimal format strings should not be exposed to UI. Only Java programmers would know what to input there without any explanation.

eselmeister:
The current change also produces artefacts, e.g. when zooming in:

Screenshot From 2025-05-07 14-45-10

Mailaender:
Added an optional decimal for zooming:
grafik

PR #2183, opened 2025-05-07
Remove old slf4.api bundle

akurtakov:
Eclipse Platform migrate to v2 long ago and it looks like nothing requires the old api in chemclipse too.


Mailaender:
https://mvnrepository.com/artifact/org.apache.jena/jena-base/3.16.0 needs it. I already tried to remove it.

akurtakov:
Closing this one as it requires far more effort for which I'm not prepared yet.

PR #2205, opened 2025-05-15
Revert "Fixed #2204 - mzML - export converter title"

### LINKED ISSUE #2204: mzML - export converter title
Display `mzML Chromatogram` instead of `MSD Chromatogram`

Mailaender:
This reverts commit 2a27003d8b1eacd94b33f2794e4fce88e7c04607. We need to distinguish MSD from HPLC-DAD here.


Mailaender:
I don't have a better name, either.

PR #2212, opened 2025-05-17
Remove the CI profile

Mailaender:
The workaround seems to be no longer required as `verify` already does not materialize products.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,4 +44,4 @@ jobs:
         MAVEN_ARGS: "--batch-mode --no-transfer-progress -Dstyle.color=always"
       run: |
         source $HOME/.sdkman/bin/sdkman-init.sh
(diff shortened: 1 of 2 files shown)


akurtakov:
https://github.com/eclipse-chemclipse/chemclipse/pull/2209 changed it but in a way that it should be run in previous stage (aka verify) thus this change should make zips/tarballs be generated again. I'll check again on Monday whether this is the case.

akurtakov:
Alternatively the gha can run "test" instead of "verify" (if it's known that nothing else bound to package phase is needed) as an addition to this change and zips/tarballs should no longer be built. See https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html for description of Maven lifecycles.

Mailaender:
`test` seems to skip the execution of the unit tests.

akurtakov:
So keeping the profile in its current way (or alternatively having the product added to the build via explicit enabling of profile) will have to be kept.

Mailaender:
I think I initially created it to suppress the tarball creation during `install`.

PR #2214, opened 2025-05-19
Make the control always print itself

### LINKED ISSUE #1994: Create Snapshot is all white on Wayland using KDE
:camera: button on the toolbar creates an empty image.

akurtakov:
Fixes https://github.com/eclipse-chemclipse/chemclipse/issues/1994 . It's best to be explicit what needs to be done.


akurtakov:
@Mailaender I would appreciate thorough testing on all your environments even if I am pretty sure this is the correct thing to do.

Mailaender (commented):
Works as promised.

PR #2218, opened 2025-05-20
Fail on compile warning

akurtakov:
Enabled unconditionally so builds fail even locally when there is a warning.


Mailaender (approved):

akurtakov:
If that's the path you would like to see forward let me know and I'll remove the failure added for demonstration purposes.

PR #2223, opened 2025-05-21
Fix jenkins warning

akurtakov:
The following warning "[WARNING] label option is deprecated. To use a static pod template, use the 'inheritFrom' option." is due to the fact that a custom image is defined (via the kubernetes section). As nothing is changed for the image it's better to use it directly.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,8 +1,6 @@
 pipeline {
 	agent {
-		kubernetes {
-			label 'ubuntu-latest'
-		}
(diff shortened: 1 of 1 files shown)


Mailaender:
See https://github.com/eclipse-chemclipse/chemclipse/pull/1953 as it does not work on Eclipse CI end.

Mailaender:
Oh, wait your change is different from https://github.com/eclipse-chemclipse/chemclipse/pull/1952.

PR #2247, opened 2025-06-07
Fix GitHub Actions tool installation

Mailaender:
@SDKMan tried to install itself twice and failed the second time. It also failed at installing what it thinks is the latest Maven: 3.9.9 while 3.9.10 is released.


Mailaender:
https://www.eclipsestatus.io/ looks like the authentication service is down again.
> Unexpected error when handling authentication request to identity provider.

PR #2250, opened 2025-06-10
Fetch Apache POI from Eclipse Orbit

Mailaender:
https://download.eclipse.org/oomph/simrel-orbit/2023-06/ ships the same version we do. This is an attempt to shrink the ChemClipse update site size.


chemclipse/releng/org.eclipse.chemclipse.targetplatform/org.eclipse.chemclipse.targetplatform.target
```diff
@@ -18,6 +18,7 @@
 	<unit id="org.apache.commons.codec"/>
 	<unit id="org.apache.commons.commons-collections4"/>
 	<unit id="org.apache.commons.commons-compress"/>
+	<unit id="org.apache.commons.io"/>
```
akurtakov: This should be org.apache.commons.commons-io and you would not have to change all the plugins.

akurtakov (approved):

Mailaender:
Included the changes from https://github.com/eclipse-chemclipse/chemclipse/pull/2252/ so we are now from 84.4 MB → 60.3 MB.

PR #2253, opened 2025-06-11
Simplify chemclipse.rcp.compilation feature

akurtakov:
Features should include only project content and fragments whenever they have are not required via p2.inf so versions are not locked. https://github.com/eclipse-simrel/.github/blob/main/wiki/SimRel/Simultaneous_Release_Requirements.md#do-not-include-non-project-content-in-features is pretty good writing on the topic.


akurtakov:
All the removed explicit requirements should be transitively resolved with the versions as specified in the bundles osgi metadata.

PR #2254, opened 2025-06-11
Add back bundles not included via osgi metadata

akurtakov:
These bundles are not included via regular OSGi metadata for some reason thus has to be explicitly required.


akurtakov:
https://github.com/eclipse-chemclipse/chemclipse/pull/2253 went too far and this one brings back 2 bundles but without "fixing" the versions.

PR #2257, opened 2025-06-11
Fetch Apache Commons CSV/Math3/XMLBeans from Orbit

Mailaender:
as we already have the repository added in https://github.com/eclipse-chemclipse/chemclipse/pull/2250. This partly reverts https://github.com/eclipse-chemclipse/chemclipse/pull/1538 as the Apache Commons CSV is a downgrade. The other versions are identical.


Mailaender:
[w=1](https://github.com/eclipse-chemclipse/chemclipse/pull/2257/files?w=1) to exclude whitespace changes.

PR #2260, opened 2025-06-11
Fix products build

akurtakov:


Mailaender (approved):
Materialize products works again.

PR #2264, opened 2025-06-12
Switch org.slf4j.api dependency to v2

akurtakov:


Mailaender:
You can then probably also remove https://github.com/eclipse-chemclipse/chemclipse/blob/7c13c16ccf17cc089ace1fc563f3774d699946c4/chemclipse/releng/org.eclipse.chemclipse.targetplatform/org.eclipse.chemclipse.targetplatform.target#L31 as it is not used anymore.

akurtakov:
Removed the ancient location in favor of using the latest slf4j.api from simrel

PR #2267, opened 2025-06-12
Downgrade to Eclipse 2025-03

Mailaender:
We do see resolver issues that cause performance problems. Wikidata test case takes 2 min. with 2025-06 and 2 seconds with 2025-03.


akurtakov:
Please refer to 2025-03 in https://github.com/eclipse-chemclipse/chemclipse/blob/7c13c16ccf17cc089ace1fc563f3774d699946c4/chemclipse/sites/chemclipse/category.xml#L17 too .

Mailaender:
Updated.

PR #2273, opened 2025-06-20
Display all ion decimals in high-resolution Tandem MS

Mailaender:
This only adds them to the scan table. I kept the scan chart as is with integers as it has less space for larger annotations.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/internal/provider/ScanLabelProvider.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/internal/provider/ScanLabelProvider.java
@@ -68,6 +68,7 @@ public class ScanLabelProvider extends ColumnLabelProvider implements ITableLabe
 	private DataType dataType;
 	//
 	private DecimalFormat decimalFormatNominalMSD;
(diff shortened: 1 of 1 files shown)


chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/internal/provider/ScanLabelProvider.java
```diff
@@ -168,9 +170,10 @@ private String getTandemMSD(Object element, int columnIndex) {
 		if(element instanceof IIon ion) {
 			IIonTransition ionTransition = ion.getIonTransition();
 			switch(columnIndex) {
-				case 0: // m/z (normal 28.3 or with Transition 128 > 78.4)
+				case 0:
 					String mz = decimalFormatTandemMSD.format(ion.getIon());
-					text = (ionTransition == null) ? mz : Integer.toString((int)ionTransition.getQ1StartIon()) + " > " + mz;
+					String q1 = decimalFormatPrecursorMSD.format(ionTransition.getQ1StartIon());
```
eselmeister: If `ionTransition` is `null`, it would crash here.

Mailaender: True

PR #2282, opened 2025-06-26
Update EJML to 0.44.0

Mailaender:
https://github.com/lessthanoptimal/ejml/blob/v0.44.0/change.txt#L5-L20


eselmeister (approved):
OK, but we need to have an eye on the test cases and RCPTT to watch if the new version has side effects.

PR #2285, opened 2025-06-26
Fix PMD violations

Mailaender:
> Unnecessary qualifier ... is already in scope because it is declared in an enclosing type


akurtakov:
Activator would better be removed entirely. The single usage of it is to get bundle version in Settings.java which can be done via `FrameworkUtil.getBundle(Settings.class).getVersion();`

akurtakov (approved):

PR #2286, opened 2025-06-26
Remove OLE32 based Office connectors

Mailaender:
This avoids Windows specific code in an otherwise multi-platform code base. Also avoids multiple versions of the same library in the platform, causing potential conflicts at runtime.


eselmeister:
You're right, it was a test to see if LibreOffice can be displayed via OLE inside of ChemClipse. It can be removed.

PR #2292, opened 2025-07-02
Remove runtime requirement on org.osgi.service.component.annotations

akurtakov:
These are needed only at build time for DS xml files generation and both Tycho and PDE will handle setting it up proper without needless explicit runtime dependency.


Mailaender:
Please rebase upon https://github.com/eclipse-chemclipse/chemclipse/pull/2293.

PR #2299, opened 2025-07-04
Include optional dependencies in the p2 repo.

akurtakov:


Mailaender:
This does not fix https://github.com/eclipse-chemclipse/chemclipse/issues/2300.

PR #2302, opened 2025-07-05
Make msd.converter.supplier test catch dependency problems too

akurtakov:
Test fragment shouldn't add extra dependencies that are needed at runtime but not for the sake of writing the test case itself. That should help in catching dependency resolving problems (at least in some cases as what PDE resolves from target platform and what is in the product can still differe significantly).
Changed the test to now swallow the exception (via assertTrue("Exception", false);) but rather let the test method throw the exception so JUnit gives the whole stacktrace in the case of assertion failure.
Migrated to JUnit 4 while at it.
Driven by https://github.com/eclipse-chemclipse/chemclipse/issues/2300 not being catched by the test.


Mailaender:
https://github.com/eclipse-chemclipse/chemclipse/pull/2301 does not fix this.

PR #2331, opened 2025-08-01
Fix Asset Installer

Mailaender:
It currently does not enable because `ConfigReader` stays null.


akurtakov (approved):
Doesn't look used at all and makes the dialog show.

PR #2333, opened 2025-08-01
Fix missing console preference page dependency

Mailaender:
This resolves the warning:
> Invalid preference category path: org.eclipse.debug.ui.ConsolePreferencePage (bundle: org.eclipse.ui.console, page: org.eclipse.ui.internal.console.ansi.preferences.AnsiConsolePreferencePage)


akurtakov (approved):
Confirm, it fixes the startup warking. 
As the preference page doesn't seem to be visible (at least in Windows/Preferences dialog) there is no change I have found at runtime.

PR #2334, opened 2025-08-01
Fixed help button not always displaying content

Mailaender:
Its reliance on the dynamic help (currently selected UI element) was twitchy with it not working properly most of the time. This now always loads the context we want when pressing the :grey_question: button.


akurtakov (approved):
Being explicit about the help topic to be displayed does no harm. 
I didn't manage to reproduce showing "wrong"/"no" content in order to check what/why is breaking.

PR #2336, opened 2025-08-01
Set focus on file item so the whole dialog can be finished via keyboard

Mailaender:
https://github.com/eclipse-chemclipse/chemclipse/pull/2150 worked on Linux with @KDE so you can hit <kbd>⏎</kbd> and finish a settings dialog with keyboard alone. This is now confirmed to work on Windows.


akurtakov (approved):
Fixes the problem on windows. It looks like SWT on Windows doesn't handle default button like on Gtk.

PR #2337, opened 2025-08-05
Don't persist the session subtract mass spectrum

Mailaender:
This avoids I/O when initializing the model bundle. Furthermore, it won't display data from previous sessions when reloading the application.


eselmeister:
This PR is problematic, due to the following reasons:

* The subtract mass spectrum field has been removed from the settings, hence the filter can't be used in a flexible way any longer.
* A defined subtract mass spectrum is not persisted any longer, hence the user needs to define and set it after restart.
* If I/O is the problem on initialization of the model bundle, then better use PreferenceSupplier.loadSessionSubtractMassSpectrum() in the UI bundle or dynamically before use.

PR #2340, opened 2025-08-07
Update binary converters to JUnit 4

Mailaender:
Same us https://github.com/OpenChrom/openchrom/pull/681 with the prospect of updating this large battery of test cases to JUnit 5 later to run them in parallel because they account for a large chunk of build time.


akurtakov (approved):
Things look good.

PR #2348, opened 2025-08-09
Port model tests to JUnit 4.

Mailaender:
Part of the https://github.com/eclipse-chemclipse/chemclipse/pull/2340 series. Introduces the new `@Test(expected = Exception.class)` syntax. Also fixes a faulty test
```patch
-		assertEquals(418731, 13f, signals.getMaxSignal());
+		assertEquals(418731.12f, signals.getMaxSignal(), 0);
```
which compared a decimal place typo with the actual value as delta.


akurtakov:
FWIW, @Test(expected...) is removed in JUnit 5 so it's not good idea to start using it now. The JUnit 4 and 5 way is using assertThrows(). See https://docs.junit.org/current/user-guide/#migrating-from-junit4-tips .

Mailaender:
Replaced it with `assertThrows(Exception.class, () -> { });` instead.

akurtakov (approved):

PR #2366, opened 2025-08-14
Fixed #2365 - Retention Index Selector - options type/delete

### LINKED ISSUE #2365: Retention Index Selector - options type/delete
Offer additional options to filter scan/peak retention index values. The following issues shall be address:

- Match Partly - currently both DB1 (non-polar) and DB1701 (semi-polar) are matched by the column name "DB1" > differentiate
- Mean, Median, ... - let the user select the option which value to use
- Fallback Column Type - in a case that DB1 is not available, use a fallback column type, to e.g. also match DB5 (non-polar)
- Delete Unrelated Indices - delete all non-matched referenced column indices

eselmeister:


chemclipse/plugins/org.eclipse.chemclipse.chromatogram.xxd.filter.supplier.scan/src/org/eclipse/chemclipse/chromatogram/xxd/filter/supplier/scan/settings/FilterSettingsRetentionIndexSelector.java
```diff
@@ -31,6 +33,18 @@ public class FilterSettingsRetentionIndexSelector extends AbstractChromatogramFi
...
+	@JsonPropertyDescription(value = "If true, the search column must not match the column name or type completely.")
+	private boolean matchPartly = false;
+	@JsonProperty(value = "Separation Column (Fallback)", defaultValue = "DEFAULT")
+	@JsonPropertyDescription(value = "Use the following separation column if the above column can't be matched.")
+	private SeparationColumnType separationColumnTypeFallback = SeparationColumnType.DEFAULT;
+	@JsonProperty(value = "Retention Index Option", defaultValue = "AUTO")
+	@JsonPropertyDescription(value = "Use the following value when assigning the retention index value.")
+	private RetentionIndexOption retentionIndexOption = RetentionIndexOption.AUTO;
+	@JsonProperty(value = "Delete Unrelated Indices", defaultValue = "false")
+	@JsonPropertyDescription(value = "Delete all indices, wich are not matched by the above settings.")
```
Mailaender, line 46: Typo

PR #2388, opened 2025-08-22
Get rid of useless activators

akurtakov:


akurtakov:
There are more like these but it's good to limit the PR to reviewable size.

Mailaender (approved):

PR #2393, opened 2025-08-25
Set user agent for BLAST and Wikidata

Mailaender:
This is enforced now by https://wikitech.wikimedia.org/wiki/Robot_policy but a good idea in general.


akurtakov (approved):
Looks good and fixes the problem.

PR #2399, opened 2025-08-26
Drop PMD and CheckStyle

Mailaender:
These rules are neither displayed nor enforced.


akurtakov (approved):
Makes sense. Having unused parts of in poms just complicate things.

PR #2407, opened 2025-08-29
Try to improve backfolding performance

Mailaender:
[BackfoldingFilter_1_ITest](https://ci.eclipse.org/chemclipse/job/chemclipse/job/develop/4215/testReport/org.eclipse.chemclipse.chromatogram.msd.filter.supplier.backfolding.core/BackfoldingFilter_1_ITest/testApplyFilter_1/history/) takes 2 minutes on the Eclipse infrastructure to run. I can't reproduce the slowness locally however I have a hunch that `addIons` is slow as in https://github.com/eclipse-chemclipse/chemclipse/pull/1900. In fact we don't need to check if the ion already exists in the mass spectrum because we just cleared it and are re-adding the backfolded ones.


Mailaender:
It had no effect on the unit test runtime speed on Eclipse CI.

PR #2412, opened 2025-09-01
Add method to determine the trace type of a given text input.

eselmeister:
Traces shall be used in a generic way, but dependent on the underlying chromatogram type, e.g. TandemMS or HighResMS might be forced to be used.


eselmeister:
The license header has been checked manually. Valid.

PR #2417, opened 2025-09-02
Fix spello

lorenzgerber:


eselmeister:
I have checked the license header manually. It's ok and valid.

PR #2428, opened 2025-09-09
PC, Fold change plot, ComboViewers

lorenzgerber:


Mailaender:
Try reloading https://github.com/eclipse-chemclipse/chemclipse/blob/develop/chemclipse/releng/org.eclipse.chemclipse.utils/codestyles/ChemClipse-Formatter.xml

PR #2448, opened 2025-09-16
Use more subtle grid chart colors in both dark and bright mode

Mailaender:
The default colors chosen in https://github.com/eclipse-swtchart/swtchart/pull/509 are too intense.


eselmeister:
"Peak Traces" still looks messy:

image

Mailaender:
Fixed `Peak Traces` and `Peak Chart`.

PR #2479, opened 2025-09-23
Remove detector slope drift

Mailaender:
Strings are objects, while enums are internalized as small integer values. It also adds type safety.


Mailaender:
Also seems to be unused except for unit tests.

PR #2483, opened 2025-09-25
Support parsing additional retention indices in *.JDX libraries.

eselmeister:


eselmeister:
Screenshot From 2025-09-26 16-28-05

PR #2490, opened 2025-09-29
Initial version of StyleText based editor

akurtakov:
Supports bold/italic/underline/striketrough text.


Mailaender (approved):
Works as promised.

PR #2511, opened 2025-10-15
Do not include LTK and Team bundles in the core feature

akurtakov:
These just end up adding extra UI elements (views/preference pages/etc.) that are totally useless in chemclipse.

### DIFF
--- a/chemclipse/features/org.eclipse.chemclipse.rcp.app.core.feature/feature.xml
+++ b/chemclipse/features/org.eclipse.chemclipse.rcp.app.core.feature/feature.xml
@@ -133,14 +133,6 @@
          version="0.0.0"/>
 
    <plugin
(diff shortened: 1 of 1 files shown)


Mailaender:
I tried to remove it before. https://github.com/eclipse-chemclipse/chemclipse/pull/1730 Not sure why it came back.

akurtakov:
AFAICT you have removed it from 2 features but missed this one.

akurtakov:
So the reason these stay is due to "Project Explorer" (https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.navigator.resources/META-INF/MANIFEST.MF) it brings ltk bundles and they bring team. I don't think we will manage to get rid of these deps and keep the project explorer.

PR #2519, opened 2025-10-20
Improve pixelated "search" icon

akurtakov:
On a hidpi screen icon used is really bad and pixelated. Use the new
https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui/icons/full/etool16/search.svg that Eclipse IDE uses to improve that.
It's put in the folder (16x16) where the search.gif was to prevent changing many other places for that.


akurtakov:
Before: 
image
After:
image

akurtakov:
Change in dark theme is even more visible. 
Before:
image
After:
image

PR #2522, opened 2025-10-22
Fix pixelated info icon

akurtakov:
Use https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface/icons/full/message_info.svg instead of the badly scaling gif version.


akurtakov:
Before: 
Screenshot From 2025-10-20 18-21-47
After:
Screenshot From 2025-10-22 15-07-08

PR #2541, opened 2025-11-05
Add support for GC-FID in JCAMP-DX format with (X++(Y..Y)) data table

Mailaender:
* [Specification](http://www.jcamp-dx.org/drafts/Chromatography%20&%20MS%20JCAMP-DX%206%20-%20Draft%2031%20May%202005.pdf)
* [Example Files](http://wwwchem.uwimona.edu.jm/spectra/index.html)


eselmeister:
Eclipse services are down again. No license check is required here.

PR #2557, opened 2025-11-17
Run license vetting only on target file change

akurtakov:
That will make some changes be found later e.g. Maven dependency defining it's dependency version with a range and new version of this dependency of dependency still within the range be pushed to Maven central.
It is needed due to EF dash tool not being able to survive the load.


Mailaender:
I think that is a good compromise. See https://github.com/eclipse-dash/dash-licenses/issues/448 for the upstream issue.

PR #2566, opened 2025-11-21
Don't round ions in Combined Scan

### LINKED ISSUE #417: Combined Scan should show all decimals
Combined mass spectrum tool works only for nominal masses... it would be awesome to implement it in order to retain all the decimal digits of the original data (independently from how many they are).

Mailaender:
Closes https://github.com/eclipse-chemclipse/chemclipse/issues/417


chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/support/CombinedMassSpectrumCalculator.java
```diff
@@ -74,21 +72,28 @@ public void addIons(List<IIon> ions, IMarkedIons excludedIons) {
...
 
 	public void removeIons(IMarkedIons excludedIons) {
 
-		for(Integer ion : excludedIons.getIonsNominal()) {
-			combinedMassSpectrum.remove(ion);
+		if(excludedIons == null) {
+			return;
+		}
+		Set<Integer> excludedIonsNominal = excludedIons.getIonsNominal();
+		java.util.List<Double> mzs = new java.util.ArrayList<>(combinedMassSpectrum.keySet());
```
eselmeister: Why don't you reference `java.util.List` and `java.util.ArrayList` via the import declaration?

chemclipse/tests/org.eclipse.chemclipse.msd.model.fragment.test/src/org/eclipse/chemclipse/msd/model/support/CombinedMassSpectrumCalculator_1_Test.java
```diff
@@ -53,13 +53,9 @@ public void testSize_4() {
...
 	public void testSize_5() {
 
-		/*
-		 * Math round is used to determine the integer value of the mass
-		 * fragment.
-		 */
 		combinedMassSpectrumCalculator.addIon(56.4f, 500.0f);
 		combinedMassSpectrumCalculator.addIon(56.2f, 700.0f);
 		combinedMassSpectrumCalculator.addIon(55.9f, 800.0f);
-		assertEquals("Size", 1, combinedMassSpectrumCalculator.size());
```
eselmeister, line 63: Add test cases for both nominal MS and high resolution MS.

chemclipse/tests/org.eclipse.chemclipse.msd.model.fragment.test/src/org/eclipse/chemclipse/msd/model/support/CombinedNominalMassSpectrumCalculator_2_Test.java
```diff
@@ -76,14 +76,10 @@ public void testSize_4() {
 	@Test
 	public void testSize_5() {
 
-		/*
-		 * Math round is used to determine the integer value of the mass
-		 * fragment.
-		 */
 		ions.add(new Ion(56.4f, 500.0f));
 		ions.add(new Ion(56.2f, 700.0f));
 		ions.add(new Ion(55.9f, 800.0f));
```
eselmeister, line 85: Watch out the intensity, when using nominal option, the final intensity shall be 2000.

chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/support/CombinedMassSpectrumCalculator.java
```diff
@@ -96,7 +96,7 @@ public ICombinedMassSpectrum createMassSpectrum(CalculationType calculationType)
 		for(Double ion : combinedMassSpectrum.keySet()) {
 			float intensity = (float)getAbundance(ion, calculationType);
 			if(intensity > IIon.ZERO_INTENSITY) {
-				massSpectrum.addIon(new Ion(ion, intensity));
+				massSpectrum.addIon(new Ion(ion, intensity), false);
```
eselmeister: That should be ok here as distinct ions are used. Watch out: if the same ion with a different intensity is added, the previous intensity will be replaced.

eselmeister:
By default, the Combined Scan support is used to create a nominal mass spectrum. Extending it to high resolution data sounds good. But I would recommend to introduce an option to retrieve either a nominal or high resolution mass spectrum. The option `nominal` shall be set by default to ensure backward compatibility.

Mailaender:
Instead of replacing the nominal mass spectrum calculator, I created a new one for high-res and a common interface for both.

PR #2577, opened 2025-11-27
Move IAnnotationWidgetService and add FileHeaderData settings UI service

eselmeister:


eselmeister:
Screenshot From 2025-11-27 09-20-23

PR #2578, opened 2025-11-27
Do not use e3 api to reset perspective

### LINKED ISSUE #901: IChemClipseEditors are destroyed when the perspective is reset
This does not happen with the 3.X editors.

akurtakov:
It doesn't know about e4 parts and thus closes the editors. Change the Main perspective views to be fully defined in e4 xmi rather than partially through code.
Fixes https://github.com/eclipse-chemclipse/chemclipse/issues/901 with the limitation that if a view has been moved it stays in the new position. If the view has been closed it's restored in its proper place.


Mailaender:
This does not seem to reset anything and it breaks the initial state of the toolbar toggles:
grafik

akurtakov:
Clarification for myself: Show/Hide is not in sync with visibility of views

akurtakov:
Defining placeholders for initial positioning will conflict with the ability for user defined positioning of views too
image

Closing as another solution will have to be found.

PR #2584, opened 2025-11-30
Replace "Show/Hide" in views controls with checkbox

akurtakov:
There is no need to compute the label each time and makes it easier to understand what happens as Show and Hide have the same width thus get easily mixed when taking a quick look


akurtakov:
Before:
image

After:
image

Mailaender:
Icon plus checkbox somehow looks broken.

akurtakov:
What do you mean? Not a clear UX or broken rendering ?

Mailaender:
It looks unpolished and inconsistent with the rest of the UI.

akurtakov:
Should I close this PR ?

eselmeister:
Screenshot from 2025-12-02 17-53-48

IMHO, it's a clearer approach, displaying the checked symbol than using "Show" or "Hide" as a prefix.

Mailaender:
It looks better on Windows.
grafik

PR #2587, opened 2025-12-02
Replace question/unknown gif with svg version from platform

akurtakov:


akurtakov:
Before: 
image

After: 
image

PR #2588, opened 2025-12-02
Replace configure.gif from svg

akurtakov:
Taken from
https://github.com/eclipse-pde/eclipse.pde/blob/6a28c28a41ebf2634f1ef960283cfb0014491d76/ui/org.eclipse.pde.ui/icons/obj16/settings.svg


akurtakov:
Before:
image

After:
image

PR #2594, opened 2025-12-03
Remove spacings between parts

Mailaender:
This might be controversial or a matter of taste but there is a general design rule that screen space should not be wasted and when compared to the Eclipse IDE our parts have a lot of space between them. Removing that and it looks like this:
grafik


akurtakov:
The extra spacing just adds noise and misaligning in my opinion. If spacing is wanted it would better be done internally in the views. +1 from me.
Old looks below for comparison sake:
image and 
image

PR #2596, opened 2025-12-04
Add optional prefixes when searching MS databases

Mailaender:
Also fixes a bug where you couldn't search for solvent:

https://github.com/eclipse-chemclipse/chemclipse/blob/b6342e73b070ed2b0f26361b1a0ed40e173279d2/chemclipse/plugins/org.eclipse.chemclipse.model/src/org/eclipse/chemclipse/model/support/LibraryInformationSupport.java#L151-L158

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.model/src/org/eclipse/chemclipse/model/support/LibraryInformationSupport.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.model/src/org/eclipse/chemclipse/model/support/LibraryInformationSupport.java
@@ -13,6 +13,7 @@
 package org.eclipse.chemclipse.model.support;
 
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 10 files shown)


chemclipse/plugins/org.eclipse.chemclipse.model/src/org/eclipse/chemclipse/model/support/LibraryInformationSupport.java
```diff
@@ -21,6 +22,23 @@
...
+		ALL, //
+		NAME, //
+		REFERENCE_IDENTIFIER, //
+		FORMULA, //
+		SMILES, //
+		INCHI, //
+		INCHI_KEY, //
+		CAS, //
+		COMMENTS, //
+		SYNONYMS, //
```
eselmeister: For the automation in `SearchField mapPrefixToField(String prefix) `or when trying to activate content proposals, wouldn't it make sense to handle the keywords in the enum?

`SYNONYMS(Arrays.asList("synonym", "synonyms")), //`

```
private List<String> keywords;
public SearchField(List<String> keywords) {
	this.keywords = keywords;
}

public List<String> keywords() {
	return keywords;
}
```

Mailaender: Moved the keywords into the enum. Also made the matching case insensitive.

PR #2612, opened 2025-12-17
Improve readability

akurtakov:
By using:
* switch expressions
* <>
* create Arrays using a[]::new


Mailaender:
Everybody loves syntax sugar!

PR #2615, opened 2025-12-18
PCA, add preference for long format import, main data file

lorenzgerber:


chemclipse/plugins/org.eclipse.chemclipse.xxd.process.supplier.pca.ui/src/org/eclipse/chemclipse/xxd/process/supplier/pca/ui/preferences/PreferencePage.java
```diff
@@ -48,5 +50,13 @@ protected void createFieldEditors() {
 		addField(new BooleanFieldEditor(PreferenceSupplier.P_CROSS_VALIDATION, "Run Cross-validation", getFieldEditorParent()));
 		addField(new ComboFieldEditor(PreferenceSupplier.P_LABEL_OPTION_PCA, "Label Option", LabelOptionPCA.getOptions(), getFieldEditorParent()));
 		addField(new ComboFieldEditor(PreferenceSupplier.P_COLOR_SCHEME, "Color Scheme", Colors.getAvailableColorSchemes(), getFieldEditorParent()));
+
+		addField(new SpacerFieldEditor(getFieldEditorParent()));
+		addField(new FileFieldEditor(PreferenceSupplier.P_MAIN_DATA_FILE, "Main Data File for Long Data import", getFieldEditorParent()) {
+
+			{
+				setFileExtensions(new String[]{"pdl"});
```
eselmeister: I assume `*.pdl` should be used here. The `FileFieldEditor` uses the following `FileDialog` method internally:
`dialog.setFilterExtensions(extensions);`
which requires to set either `*.*` to match all or specific extensions like `*.pdl`.

PR #2626, opened 2026-01-06
Mark constant values as static final 

Mailaender:
This is probably more about clarity than performance.


akurtakov:
With https://openjdk.org/jeps/500 it might become about performance too.

PR #2639, opened 2026-01-14
Avoid PathResolver duplication

Mailaender:
Every bundle had their own `PathResolver` class, because its implementation relied on using the current bundle, which led to a lot of redundancy. We can achieve the same with just one class by taking the bundle that shall be searched as input. Also, I/O errors are now logged instead of printed to console.


eselmeister:
The PR looks reasonable. We should watch the cases, where e.g. the absolute path to a directory or file is requested, that doesn't exist yet. For example in a testing environment: create a report that doesn't exist yet, run reporting and test the file. Same for directory. Solution here: request the parent directory that exists and then create a new file that represents the needed directory or file.

Mailaender:
You probably didn't see it because the diff was too large, but this is how it is done https://github.com/eclipse-chemclipse/chemclipse/blob/c919060c7f180365064ee45642aa90678bc11e7e/chemclipse/tests/org.eclipse.chemclipse.msd.converter.supplier.amdis.fragment.test/src/org/eclipse/chemclipse/msd/converter/supplier/amdis/converter/msl/MSLExportConverter_1_ITest.java#L48-L49 Use a path that exists and then append to it.

eselmeister:
Perfect, I was sure that you had this in mind already. Thanks for taking care of it.

PR #2645, opened 2026-01-16
When trying to resolve the alkanes.msl file via UI, use class from model

eselmeister:


Mailaender (approved):
Ah, that makes sense.

PR #2647, opened 2026-01-19
Remove standard wizards by default to clean up the UI.

eselmeister:


eselmeister:
@akurtakov Please also review and comment.

akurtakov (approved):
Totally makes sense to be sure whether wizards are or aren't there to have less surprises at runtime depending on how someone started his app.

Mailaender:
Context: https://github.com/eclipse-chemclipse/chemclipse/issues/1266

PR #2662, opened 2026-01-28
Add a filter to nominalize an MSD chromatogram an its references.

eselmeister:


eselmeister:
We already have a "Scan Mass Spectrum" filter to nominalize m/z values to unit mass, but the chromatogram context is not available there. Hence, it is not possible to also nominalize reference chromatograms. It's needed e.g. if a scan density filter shall be applied on the master and its references after nominalization.

eselmeister:
Screenshot From 2026-01-28 08-39-29
Screenshot From 2026-01-28 08-39-36

PR #2672, opened 2026-02-04
Replace console view with error log

Mailaender:
I removed the Eclipse debug dependency to reduce the number of hotkeys (and conflicts), which then called for an error message in the log view, which has its preference page tied to debugging and it is also a part of the platform debugging tools. All of its messages except for the folders are redundant to the feedback view and status line logger.

The error log, despite its name, is not part of PDE but actually part of the more general `eclipse.platform.ui` package. Before it was only accessible through the hidden hotkey <kbd>CTRL</kbd> + <kbd>F3</kbd> and searching for it, despite actually being very helpful when tracking down problems.


akurtakov:
I really like the reduction of shipped bundles coming from this change. How useful Console plugin is for existing users is smth I can not evaluate though.

eselmeister:
I used the console years ago to compile Java and Groovy code inside of OpenChrom. Well, that was one of those experiments.

PR #2698, opened 2026-02-10
Type the peak model interfaces

Mailaender:
I believe we can move the cast, which is always safe, to the interface.


eselmeister:
Looks reasonable.

PR #849, opened 2021-12-15
Added default icons to the quick access toolbar

Mailaender:
Close https://github.com/eclipse/chemclipse/issues/848

PR #2701, opened 2026-02-11
Harmonize processor icon toolbar icons

Mailaender:
This now re-uses the icon definitions from #1093 which are not available back in #849. This means downstream plug-ins can contribute their own icons. You can still override them with custom ChemClipse platform icon like before. I also fixed missing icons when using the German localization.


chemclipse/plugins/org.eclipse.chemclipse.chromatogram.xxd.quantitation.supplier.chemclipse.ui/plugin.xml
```diff
@@ -5,7 +5,7 @@
          point="org.eclipse.chemclipse.xxd.process.ui.menu.icon">
       <icon
             class="org.eclipse.chemclipse.chromatogram.xxd.quantitation.supplier.chemclipse.ui.icon.InternalStandardsMenuIcon"
-            id="org.eclipse.chemclipse.chromatogram.xxd.quantitation.supplier.chemclipse.peak.istd">
+            id="org.eclipse.chemclipse.chromatogram.msd.quantitation.supplier.chemclipse.peak.istd">
```
eselmeister: ISTD peak quantification can be also applied on GC-FID.

Mailaender: Okay, I changed it the other way round. Note that this is an ID change that requires an upgrade rule that I proposed in https://github.com/eclipse-chemclipse/chemclipse/pull/2124.

PR #2704, opened 2026-02-11
Add mass spectra centroidisation

Mailaender:
This converts profile to centroided spectra based on detected peaks. Oftentimes this process means peak picking and then converting the spectra in one go, but I want to separate the two as we may want to try multiple peak picking algorithms. I also want people to preview the picked peaks on the profile spectrum to judge the result. Centroidization will then also be an additional, destructive, shared step. We already have process methods to orchestrate everything into one again.


chemclipse/plugins/org.eclipse.chemclipse.msd.filter.supplier.centroiding/src/org/eclipse/chemclipse/msd/filter/supplier/centroiding/core/MassSpectrumFilter.java
```diff
@@ -0,0 +1,90 @@
...
+		return processingInfo;
+	}
+
+	private IProcessingInfo<IMassSpectrumFilterResult> centroidisation(IScanMSD scanMSD, MassSpectrumFilterSettings massSpectrumFilterSettings) {
+
+		IProcessingInfo<IMassSpectrumFilterResult> processingInfo = validate(scanMSD, massSpectrumFilterSettings);
+		if(processingInfo.hasErrorMessages()) {
+			return processingInfo;
+		}
+		if(scanMSD instanceof IStandaloneMassSpectrum massSpectrum) {
```
eselmeister, line 58: Why don't you use `XPassFilter.nominalize(...);` here?

chemclipse/plugins/org.eclipse.chemclipse.msd.filter.supplier.centroiding/src/org/eclipse/chemclipse/msd/filter/supplier/centroiding/core/MassSpectrumFilter.java
```diff
@@ -0,0 +1,90 @@
...
+			processingInfo = validate(massSpectrum);
+			if(processingInfo.hasErrorMessages()) {
+				return processingInfo;
+			}
+			massSpectrum.removeAllIons();
+			for(IMassSpectrumPeak massSpectrumPeak : massSpectrum.getPeaks()) {
+				IIon ion = new Ion(massSpectrumPeak.getIon(), (float)massSpectrumPeak.getAbundance());
```
eselmeister, line 65: The above `XPassFilter` does the trick already. If not used, `getIon()` could return a double with decimals here, that needs to be casted to Integer e.g. via `AbstractIon.getIon(mz)`. But then it needs to be checked, that the intensities of e.g. 56.12 and 56.16 are summed.

chemclipse/plugins/org.eclipse.chemclipse.msd.filter.supplier.centroiding/src/org/eclipse/chemclipse/msd/filter/supplier/centroiding/core/MassSpectrumFilter.java
Mailaender: Because this is designed for MALDI-ToF MS. Nominalizing would remove all the precision. Also, if you want to nominalize beforehand, you could just run that filter separately. It does not make sense to combine the two.

chemclipse/plugins/org.eclipse.chemclipse.msd.filter.supplier.centroiding/src/org/eclipse/chemclipse/msd/filter/supplier/centroiding/core/MassSpectrumFilter.java
Mailaender: I don't want to round ions here. Just convert profile to centroid. The centroided one can then be used for identification next step.

eselmeister:
OK, I got it.

PR #2705, opened 2026-02-12
Fix loading of SVGs at different size

akurtakov:
It's a limitation in how SWT treats SVG images - uses the width/height in the svg file and treats it as fixed size to fit into all the use cases with old APIs.
Use SWT's internals to load at requested size and store in the cache with key that includes the size.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.rcp.ui.icons/src/org/eclipse/chemclipse/rcp/ui/icons/core/AbstractApplicationImage.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.rcp.ui.icons/src/org/eclipse/chemclipse/rcp/ui/icons/core/AbstractApplicationImage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013, 2025 Lablicate GmbH.
+ * Copyright (c) 2013, 2026 Lablicate GmbH.
  *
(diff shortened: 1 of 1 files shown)


akurtakov:
If I suppress the warnings with annotation for the IDE Maven builds fails with:
```
[WARNING] /home/akurtakov/git/chemclipse/chemclipse/plugins/org.eclipse.chemclipse.rcp.ui.icons/src/org/eclipse/chemclipse/rcp/ui/icons/core/AbstractApplicationImage.java:[63] 
	@SuppressWarnings("restriction")
	                  ^^^^^^^^^^^^^
Unnecessary @SuppressWarnings("restriction")
[WARNING] /home/akurtakov/git/chemclipse/chemclipse/plugins/org.eclipse.chemclipse.rcp.ui.icons/src/org/eclipse/chemclipse/rcp/ui/icons/core/AbstractApplicationImage.java:[70] 
	@SuppressWarnings("restriction")
	                  ^^^^^^^^^^^^^
Unnecessary @SuppressWarnings("restriction")
2 problems (2 warnings)
error: warnings found and -failOnWarning specified
```

akurtakov:
So this is a limitation in Tycho as it doesn't apply all the analysis pde does. Options are:
* disable -failOnWarning in Tycho
* accept warning in workspace
* do some non trivial work on tycho to support it

PR #2706, opened 2026-02-12
Avoid authorship with nicknames in JavaDoc

Mailaender:
We already have full names in :copyright: headers and Git commits that tie developers to all the lines changed. Using nicknames in the official documentation seems awkward given that this is a commercial open-source project. Also, only a select few people used it on a few occasions.


eselmeister:
Yes, good idea. It historical code and was auto-generated.

PR #2727, opened 2026-02-17
Use getDisplay() instead of getting it from the event.

eselmeister:


chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/swt/ExtendedPeakScanListUI.java
```diff
@@ -475,7 +475,7 @@ public void handleEvent(ExtendedTableViewer extendedTableViewer, KeyEvent e) {
 				} else if(matchesKeyPress("Unknown", e)) {
 					addTargetsUnknown(e.display);
```
Mailaender, line 476: Maybe here as well. Sounds like a good idea in general.

eselmeister: Yep, that seems to work. I plan to make a general refactoring.

PR #2757, opened 2026-03-01
Enable to set a target in the scan/peak list via the clipboard.

eselmeister:


eselmeister:
Use `Copy & Paste` from the clipboard to set a target.

Screenshot From 2026-03-01 10-23-23
Screenshot From 2026-03-01 10-23-34

PR #2764, opened 2026-03-02
Remove the Peak Identification Batch Editor

Mailaender:
I believe it is completely redundant to the more flexible Batch Job UI, which can also do peak integration, identification and saving of the peaks as text file.


eselmeister:
This batch job was designed for peak input files, that are decoupled from a chromatogram. So you have a list of GC-MS peaks and would like to integrate and identify them. But yes, it's old code and can be removed to clean up the platform.

PR #2779, opened 2026-03-04
EditorCalibration shouldn't be multi page

akurtakov:
It is single page so it just removes single ctab at the bottom.


akurtakov:
Before: 
Screenshot From 2026-03-04 16-36-44
After:
Screenshot From 2026-03-04 16-34-58

PR #2784, opened 2026-03-05
NMRBatchJob inject instead of static access

akurtakov:
Use injected ProcessingInfoPartSupport and IEclipseContext instead of getting them from Activator.


eselmeister:
@akurtakov I will test it right now.

PR #2796, opened 2026-03-15
Use ChromatogramEditorMSD partDescriptor in demo loading

akurtakov:
Create partDescriptor in fragment.e4xmi and use it.


akurtakov:
The pattern should expand to all editors but changing it in SupplierEditorSupport is not straightforward thus landing this first piece in here.

PR #2797, opened 2026-03-15
Simplify ISupplierFileEditorSupport

akurtakov:
* Remove useless label - in a single cycle break is enough
* Do not manually play with activeChildContext - this is smth that should be managed by e4 based on the model. Also it seems to have no immediate effect.


Mailaender:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=502544 is marked as resolved now.

PR #2799, opened 2026-03-16
Update to 2026-03 Eclipse simrel

akurtakov:


akurtakov:
License check failure is to be ignored as EF service just errors now: "Error:  Error response from the Eclipse Foundation 504"

PR #2800, opened 2026-03-16
Fix Tried to access shell outside of UI thread! exception.

eselmeister:


chemclipse/plugins/org.eclipse.chemclipse.chromatogram.filter.ui/src/org/eclipse/chemclipse/chromatogram/filter/ui/core/ChromatogramFilter.java
```diff
@@ -60,14 +60,15 @@ public IProcessingInfo<IChromatogramFilterResult> applyFilter(IChromatogramSelec
...
+		 */
 		if(!processingInfo.hasErrorMessages()) {
-			Shell shell = DisplayUtils.getShell();
-			if(shell != null) {
-				detectScanMaxima(shell, chromatogramSelection, filterSettings);
-			} else {
-				DisplayUtils.getDisplay().syncExec(() -> {
-
+			Display.getDefault().asyncExec(() -> {
+				Shell shell = Display.getDefault().getActiveShell();
```
Mailaender: Try
```java
@Named(IServiceConstants.ACTIVE_SHELL)
private Shell shell;
```
instead.

akurtakov: From what I understand filters are instantiated using extension registry (element.createExecutableExtension...) and this doesn't do injection, it just creates the object . In order for injection to happen ContextInjectionFactory.inject()  will have to be called wherever instantiating. Alternatively ContextInjectionFactory.make() could be tried to do instantiating but that would most probably conflict with setting the properties from the extension.

eselmeister: Correct. I updated to the injection anyways to be future proof.

PR #2803, opened 2026-03-16
Revert "Use ChromatogramEditorMSD partDescriptor in demo loading"

akurtakov:
Reverts eclipse-chemclipse/chemclipse#2796 as the partDescriptor content seems to leak into instances created with ModelService too - aka all cases will have to be updated at once at some point.


akurtakov:
The visible bug is that opening second/third/etc. chromatogram it doesn't get auto activated.

PR #2807, opened 2026-03-17
Remove empty preference pages

Mailaender:
as they clutter the UI in both preferences and the hotkey settings.


eselmeister:
Sometimes preference pages, even empty ones, are used as top level pages for sub pages. So far, it seems that this PR only removes empty sub pages.

Mailaender:
I didn't touch the categories.

PR #2811, opened 2026-03-18
Stop the identification process when CANCEL is pressed in wizard.

eselmeister:


eselmeister:
Screenshot From 2026-03-18 15-37-03

PR #2812, opened 2026-03-19
Remove unused static methods in ChromatogramType

akurtakov:
Not used and reduces static access to IEclipseContext


Mailaender:
I believe it is not in use because the underlying e4 core expressions are not working correctly. https://github.com/eclipse-chemclipse/chemclipse/commit/f84e8210583c2a0c58db14c5db1d5d2a6fb2924f

akurtakov:
I would need some guidance/explanation about what/where these expresions are tried to be used . Having these static accesses is not the right way to do them IMO.

eselmeister:
Let me review it.

eselmeister:
Yes, the `org.eclipse.core.expressions.definitions` was used to retrieve the status of the current chromatogram selection, for instance MSD (`CHROMATOGRAM_TYPE_MSD`). In the *.e4xmi model, it was possible to attach the definitions, when the menu item shall be displayed or not via the "`Core Expression`". For menu items, it worked. But it failed for toolbar items. IMHO, we don't need this anymore, as we found other solutions, like dynamic menus and specific toolbars.

PR #2832, opened 2026-04-03
Remove ChromatogramType EventHandler

akurtakov:
EclipseContextFactory.getServiceContext(bundleContext) returns the OSGi-service-backed context for that specific bundle's registry. 
The E4 application context (MApplication.getContext()) is a completely separate object; it is a child of the platform workbench's service context, not of the org.eclipse.chemclipse.ux.extension.ui bundle's service context. 
Aka the value was set in sibling and not in parent context which leads to this evaluations in E4 were never working proper. In order to do it proper it would have to happen via E4 and not via OSGi DS.


akurtakov:
ExtendedChromatogramUI.addMainMenu is creating the menu entries manually . I would rather see OSGi DS implementation gone before trying to do it on E4 layer in order to mix less things.

PR #2834, opened 2026-04-06
Remove *Type EventHandlers

akurtakov:
Much like https://github.com/eclipse-chemclipse/chemclipse/pull/2832 these write in their own context thus unusable. Combined with the fact it is not used anywhere it can be safely removed.


chemclipse/plugins/org.eclipse.chemclipse.ux.extension.ui/META-INF/MANIFEST.MF
```diff
@@ -67,10 +67,4 @@ Import-Package: jakarta.annotation;version="2.1.1",
  jakarta.inject;version="2.0.1",
  org.osgi.framework;version="[1.0.0,2.0.0)",
  org.osgi.service.event;version="[1.0.0,2.0.0)"
```
Mailaender: Can this be removed as well?

akurtakov: ProcessMethodNotifications is still OSGi DS thus no, it can't be removed.

PR #2839, opened 2026-04-15
Add copilot-setup-steps.yml to set Java 21 for agent sandbox

akurtakov:


akurtakov:
This is an effort trying to make copilot in github be able to verify the changes it makes, which failed before this pr due to usage of Java 17 by default and later due to the bundle host mismatch (puzzles me but its easier to change it than to investigate).

PR #2856, opened 2026-04-23
Remove most e3 editor extension points

Mailaender:
Since https://github.com/eclipse-chemclipse/chemclipse/pull/2767 we no longer need these extension points except for two, which are still e3 and wrapped around for use in e4. I added them in https://github.com/eclipse-chemclipse/chemclipse/pull/1793 and https://github.com/eclipse-chemclipse/chemclipse/pull/1785/ to enable drag and drop, but that only ever worked from Project Explorer.


akurtakov:
It's a huge relief to see more of these e3 editors gone

PR #2857, opened 2026-04-23
Move chart titles from preferences to NLS

Mailaender:
They should be configurable but not in preferences, as they are natural language. This also reduces the scope of some very overloaded preference pages. Also another small step towards https://github.com/OpenChrom/openchrom/issues/24.

It seems that parent/daughter ion terminology is deprecated. I think it is precursor/product ion nowadays, but I will leave this as is for now.


eselmeister:
This removes the possibility for the user to modify the title, am I right?

Mailaender:
You could create a user language and modify the titles, including all the other strings that are already translatable.

eselmeister:
Once we had requests of users, who wanted to change the label e.g. "Retention Time [min]" to "Retention Time (min)". With the current change, can they do it on-the-fly? Or does it require extra work to ship specific language pack files?

Mailaender:
Easiest way would be to edit the existing `messages.properties`. A more robust approach would be shipping a custom language pack. We do lose an easy (GUI) override. At least I am not aware of any mechanism. :man_shrugging:

PR #2901, opened 2026-05-18
Migrate pdfbox to 3.0.7

akurtakov:
PDFBox doesn't implement equals in its model classes and there are no more constants so some adjustments to tests are needed.


eselmeister:
relates to:
https://github.com/eclipse-swtchart/swtchart/pull/586

PR #2902, opened 2026-05-19
Feature, Save and Load PCA Evaluation

lorenzgerber:
Save and load IEvaluation objects, using serialization to json

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.xxd.process.supplier.pca/src/org/eclipse/chemclipse/xxd/process/supplier/pca/io/WorkspaceIO.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.xxd.process.supplier.pca/src/org/eclipse/chemclipse/xxd/process/supplier/pca/io/WorkspaceIO.java
@@ -1,0 +1,88 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Lablicate GmbH.
+ *
+ * This program and the accompanying materials are made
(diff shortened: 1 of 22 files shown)


chemclipse/plugins/org.eclipse.chemclipse.xxd.process.supplier.pca/src/org/eclipse/chemclipse/xxd/process/supplier/pca/io/WorkspaceIO.java
```diff
@@ -0,0 +1,88 @@
...
+import org.eclipse.chemclipse.xxd.process.supplier.pca.model.IResult;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
+public class WorkspaceIO {
+
+	public static final String DESCRIPTION = "PCA Workspace";
+	public static final String FILE_EXTENSION = ".pcaws.json";
```
eselmeister: It's a JSON file. But I recommend to use a distinct name here to avoid that people try to load any *.json file, that is not of the specific PCA workspace type. Probably name it.

`public static final String FILE_EXTENSION = ".pws";`

lorenzgerber: agree

PR #2906, opened 2026-05-21
Make forbiddenReference error

akurtakov:
PDE has a new warning
eclipse-pde/eclipse.pde#2339 that shows errors in the workspace.


Mailaender:
But you are setting it to error here?

akurtakov:
"a value of the org.eclipse.jdt.core.compiler.problem.forbiddenReference preference other than error effectively disables the enforcement of almost all OSGi access rules. This potentially leads to classloading errors at runtime because access rule violations are not detected anymore at compile-time/in the IDE." 
It must be "error" , otherwise PDE (in 2026-06) will show error like 
image

PR #2919, opened 2026-06-01
Disable error log window focussing on new event

Mailaender:
because it hides the scan chart which is annoying.

grafik


akurtakov:
I believe https://github.com/eclipse-chemclipse/chemclipse/pull/2920 is better fix as it doesn't rely on IStartup (which is e3 concept).

PR #2920, opened 2026-06-01
Totally disable error log window focussing on new event

akurtakov:
LogView uses both Instance (user saved) and is nothing saved Default scopes (new workspace) see
https://github.com/eclipse-platform/eclipse.platform.ui/blob/62c792ecba2353705cfe1d7fd9a96ff603e9040f/bundles/org.eclipse.ui.views.log/src/org/eclipse/ui/internal/views/log/LogView.java#L1773 . This makes plugin_customizations have no effect on existing workspaces.
Set the values we want in ApplicationWorkbenchWindowAdvisor .


akurtakov:
In theory that should be addon but addon and e3 compat layer seem to run in the same phase so order is unpredictable. Whenever Chemclipse is pure e4 app this could be reconsidered.

PR #2933, opened 2026-06-09
Mark duplicates by name and/or CAS# and use a specific column image.

eselmeister:


eselmeister:
Screenshot From 2026-06-09 11-08-01

PR #2938, opened 2026-06-12
Update to 2026-06 simrel

akurtakov:


Mailaender:
It should also contain Apache PDFBox. https://github.com/eclipse-orbit/orbit-simrel/commit/3f1af0c82aa8317f4acdec3c3506831f1906a5b5

chemclipse/releng/org.eclipse.chemclipse.targetplatform/org.eclipse.chemclipse.targetplatform.target
```diff
@@ -58,46 +58,13 @@
 		<repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2026-06"/>
 		<unit id="org.apache.commons.commons-csv"/>
 		<unit id="org.apache.commons.math3"/>
+		<unit id="org.apache.pdfbox.preflight"/>
```
Mailaender: Preflight is a validation tool aimed at developers, not end users.

akurtakov: Do you mean it shouldn't be in the target platform?

Mailaender: Yes, I don't think we need this here. I only used the standalone `.jar` at https://github.com/eclipse-swtchart/swtchart/issues/417 once.

Mailaender:
You can also remove the workaround in https://github.com/eclipse-chemclipse/chemclipse/blob/fb1d2957bfd14ce912053c818713c811dac35512/chemclipse/plugins/org.eclipse.chemclipse.ux.extension.ui/src/org/eclipse/chemclipse/ux/extension/ui/provider/DataExplorerLabelProvider.java#L82-L87 as https://github.com/eclipse-platform/eclipse.platform.swt/pull/3262 should now be available.

akurtakov:
Let's have these extra changes after the move to 2026-06 is done .

PR #2942, opened 2026-06-16
Reset perspective on e4 model

### LINKED ISSUE #901: IChemClipseEditors are destroyed when the perspective is reset

akurtakov:
Perspectives are contributed via the e4 model in 2026-06, so the legacy IWorkbenchPage#resetPerspective() can't be used (it will NPE on the missing e3 perspective descriptor). Reset by cloning the snippet stored on startup.
Fixes https://github.com/eclipse-chemclipse/chemclipse/issues/901


Mailaender (approved):
Works as promised.

Mailaender:
Works for every perspective except PCA which will reset to the welcome state which is probably a design problem with the dynamic layout.

PR #2948, opened 2026-06-18
Dynamically edit the ComboSettingsProperty, by default edit=false

eselmeister:


eselmeister:
```
@ComboSettingsProperty(edit = true, value = ContributorComboSupplier.class)
private String specificContributor = "";
```
Screenshot From 2026-06-18 15-36-07

PR #2958, opened 2026-06-29
Fix NPE/wrong handling when supplier id is not set in the map.

eselmeister:
Introduced in:
https://github.com/eclipse-chemclipse/chemclipse/pull/2951


chemclipse/plugins/org.eclipse.chemclipse.ux.extension.xxd.ui/src/org/eclipse/chemclipse/ux/extension/xxd/ui/editors/AbstractChromatogramEditor.java
```diff
@@ -396,6 +396,22 @@ private synchronized IChromatogramSelection loadChromatogram() {
...
+	/**
+	 * Could be also null if not set yet.
+	 * 
+	 * @return String
+	 */
+	private synchronized String getSupplierID(Map<?, ?> map) {
+
+		String supplierId = null;
+		Object id = map.get(EditorSupport.MAP_SUPPLIER_ID);
+		if(id != null && !id.toString().isBlank()) {
```
Mailaender: Maybe even check for `id instanceof String` to be extra safe.

PR #2960, opened 2026-06-30
Add an option to show the flavor markers and enable to modify them.

eselmeister:


eselmeister:
Screenshot From 2026-07-01 14-24-39
Screenshot From 2026-07-01 14-24-44

PR #2961, opened 2026-07-01
Add a dialog to display the mass spectrum and let the user modify it.

eselmeister:


eselmeister:
Screenshot From 2026-07-01 14-25-18
Screenshot From 2026-07-01 14-25-22

PR #2963, opened 2026-07-02
Avoid ion transitions in MS1 spectra to save memory

Mailaender:
This is an API change with the motivation to reduce RAM usage on mass spectra. `IIon` is generally the most abundant object in both MS and LC/GC-MS. Removing ion transitions from MS1 spectra yields 25% fewer allocations. Apparently even a null object causes overhead. Measured with @visualvm. By introducing `IIonMSn` we also get type safety for tandem MS, which should be safer than null checks on optional fields.


Mailaender:
Cleaner diff: https://github.com/eclipse-chemclipse/chemclipse/pull/2963/changes?w=1 avoids whitespace noise.

eselmeister:
This PR needs further review as it changes the way how to create ion transition instances.

Mailaender:
It actually just moves code around, but it touches plenty of places.

PR #2969, opened 2026-07-04
Enable to select multiple views to display via the ViewDialog.

eselmeister:


eselmeister:
Screenshot From 2026-07-04 11-14-17
Screenshot From 2026-07-04 11-14-42

PR #2971, opened 2026-07-06
Enable a dynamic library entry duplicate detection via Name or CAS.

eselmeister:


eselmeister:
image

eselmeister:
image

eselmeister:
image

eselmeister:
image

Mailaender:
I think enabling SPLASH would make more sense here. What if there is a duplicate spectrum with a missing CAS and a different name? This is actually searching for technical replicates.

eselmeister:
I'm fine to add additionally "Splash" as an option for duplicate detection here.

PR #2973, opened 2026-07-06
Use primitive hash helpers

Mailaender:
This reduces allocations and is cleaner.


eselmeister:
The Eclipse IDE offers a function to generate the hashCode, equals and toString method. Why don't you use that one, e.g. for `RetentionIndicesSupplier`? 

```java
@Override
public int hashCode() {

	return Objects.hash(description, exportable, fileExtension, filterName, id, importable);
}

@Override
public boolean equals(Object obj) {

	if(this == obj)
		return true;
	if(obj == null)
		return false;
	if(getClass() != obj.getClass())
		return false;
	RetentionIndicesSupplier other = (RetentionIndicesSupplier)obj;
	return Objects.equals(description, other.description) && exportable == other.exportable && Objects.equals(fileExtension, other.fileExtension) && Objects.equals(filterName, other.filterName) && Objects.equals(id, other.id) && importable == other.importable;
}

@Override
public String toString() {

	return "RetentionIndicesSupplier [id=" + id + ", description=" + description + ", filterName=" + filterName + ", fileExtension=" + fileExtension + ", exportable=" + exportable + ", importable=" + importable + "]";
}
```

Mailaender:
`Objects.hash` might change behavior and is slower than the manual implementation, so better avoid it in hot paths like `AbstractIon` which is the most abundant object in the platform.

PR #2975, opened 2026-07-07
Notify scan chart after importing ions from scan proxies

Mailaender:
For indexed mzML I don't just import the ions but set up the whole metadata around the scan, which is inaccessible unless the whole `<spectrum>` that also includes the binary blob is deserialized, so I do it in the proxy:

https://github.com/eclipse-chemclipse/chemclipse/blob/3c69cbbb9373873932b40aab73af3b0c4cc9f84f/chemclipse/plugins/org.eclipse.chemclipse.xxd.converter.supplier.mzml/src/org/eclipse/chemclipse/msd/converter/supplier/mzml/converter/model/VendorScanProxy.java#L62-L64

That means after the proxy is read, the scan charts need an update to display profile instead of centroidation and show the correct polarization, MSn etc.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/core/AbstractRegularMassSpectrumProxy.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.msd.model/src/org/eclipse/chemclipse/msd/model/core/AbstractRegularMassSpectrumProxy.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2025 Lablicate GmbH.
+ * Copyright (c) 2015, 2026 Lablicate GmbH.
  *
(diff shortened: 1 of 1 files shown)


eselmeister:
In case of a loading a single scan, that's ok. If the whole chromatogram is processed, then `UpdateNotifier.update(this)` would be called n times and fires the event `IChemClipseEvents.TOPIC_SCAN_XXD_UPDATE_SELECTION`, that could lead to updating the relevant parts n times, which could slow down the performance. Did you have this scenario in mind and did you test how the system reacts?

Mailaender:
I think you are right. This will trigger too much. Only tested clicking into the chromatogram and then updating the scan chart. I consider the current scan proxy system broken anyway because when I do any action that goes through all the scans one by one, then for me the system just stalls, the progress bar does not update and everything is loaded into RAM and never unloaded as the scans keep getting referenced by the chromatogram.

PR #2979, opened 2026-07-08
Show labels in the selection dialog sorted. Refactor names to be unique.

eselmeister:


Mailaender:
I think a better approach would be: only display views of the current perspective.

eselmeister:
I'm against only showing the views of the current perspective, as this is the flexibility of the Eclipse approach to freely select the part you'd to show. If e.g. the Molecule part is only defined in "Data Analysis", then you can't load it in the PCA perspective.

Mailaender:
Then maybe group them by perspective.

eselmeister:
Yes, AFAIK ... we get that out of the box, when we switch back to the Eclipse perspective and view switcher.

PR #2983, opened 2026-07-09
Enable to only show duplicates in the mass spectrum library editor.

eselmeister:


eselmeister:
Screenshot From 2026-07-10 07-44-46
Screenshot From 2026-07-10 07-44-51

PR #2989, opened 2026-07-14
Workaround duplicate feedback part ids

akurtakov:
Several perspectives share the feedback part id, so showPart(String, PartState) is ambiguous.
Look up the part in the active perspective instead using part service comes from the window directly, not the context's active child. The message is pushed into the shown part directly since a freshly created part's own injected notifier may not be the same instance this class updated.


Mailaender (approved):
Works as promised.

PR #3011, opened 2026-07-27
Make BatchJobEditor a pure E4 part

akurtakov:
Drops BatchJobEditor's EditorPart/org.eclipse.ui.editors registration in favor of an E4 MPart (IChemClipseEditor + PostConstruct/Persist/Focus annotated methods), mirroring EditorCalibration.

Files of type .obj opening now routes through the generic E4 openEditor(...) helper, resolved via Adapters.adapt(...) on BatchJobFileSupplier. The editor is opened by BatchProcessHandler once the wizard dialog is closed, so the E4 part service can resolve the active window.


Mailaender (approved):
Batch converter still works.

PR #3014, opened 2026-07-28
Strip characters invalid in XMI from part labels

akurtakov:
The e4 workbench model is persisted as XMI, which rejects characters it is unable to represent instead of escaping them. A part label built from vendor data then fails the workbench auto save.

Sanitize where a value enters the workbench model, so that the measurement data itself stays byte faithful.


akurtakov:
@Mailaender Please verify this PR is what we spoke about. It's noisier than I hoped but I haven't found a more central place to do the sanitization.

Mailaender (approved):
Works as promised.

PR #3019, opened 2026-07-30
Reduce visual clutter in MALDI-TOF MS charts

Mailaender:
This applies https://github.com/eclipse-swtchart/swtchart/pull/601/ to the MALDI-TOF MS perspective.

**Before:**
grafik

**After:**
grafik

The idea here is fewer decimal places make it easier to read and give more room to the chart itself.


Mailaender:
In fact I can simplify this even further by removing the secondary axis:
grafik

PR #3032, opened 2026-08-10
Collect/show the invalid CAS# and let the user decide which to remove.

eselmeister:


eselmeister:
Screenshot From 2026-08-10 07-57-40

PR #3033, opened 2026-08-10
Add options formula, smiles, inchi and reference for duplicate detection

eselmeister:


eselmeister:
Screenshot From 2026-08-10 07-57-51

PR #3044, opened 2026-08-17
Remove unused function

Mailaender:
as it is in fact unused. https://github.com/eclipse-chemclipse/chemclipse/commit/d2fcecc89e39d183589bf4350db1a47609396c51 added the function without context.


Mailaender:
The activator is indirectly used by `PreferenceSupplier.getSelectedNoiseCalculatorId()`.

PR #3072, opened 2026-08-26
Add a read-only mzPeak converter

Mailaender:
This is an early version of https://github.com/HUPO-PSI/mzPeak support. The format uses Apache Parquet. It is a large project with many third-party dependencies. It requires almost all of Apache Hadoop, which is not just huge but also takes a while to initialize or does not properly work in an OSGi environment. There is an open ticket about removing those dependencies for client-side applications https://issues.apache.org/jira/browse/PARQUET-1822 but it seems to be a difficult refactor. I therefore went with @duckdb, which is just a JDBC driver that understands Parquet and is easy to work with.


Mailaender:
https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/work_items/30529

Mailaender:
> license-check failed: java.net.http.HttpTimeoutException: request timed out

I am ignoring the failure because the new dependency was just vetted.

PR #3085, opened 2026-08-31
Add a wizard to create regex definitions to extract/transfer data.

eselmeister:


eselmeister:
Screenshot From 2026-08-31 19-25-34
Screenshot From 2026-08-31 19-26-14

PR #3105, opened 2026-09-03
Add XmlParserFactory and use it for the various file formats

akurtakov:
JDK 24 capped maxGeneralEntitySizeLimit and totalEntitySizeLimit at 100.000, which large mzML files exceed via escaped characters.

### DIFF
--- a/chemclipse/plugins/org.eclipse.chemclipse.support/src/org/eclipse/chemclipse/support/xml/XmlParserFactory.java
+++ b/chemclipse/plugins/org.eclipse.chemclipse.support/src/org/eclipse/chemclipse/support/xml/XmlParserFactory.java
@@ -1,0 +1,65 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Lablicate GmbH.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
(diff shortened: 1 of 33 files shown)


akurtakov:
@Mailaender Please confirm this fixes your local tests exceeding the limits. Hopefully the factory didn't went too far with applying all settings.

chemclipse/plugins/org.eclipse.chemclipse.support/src/org/eclipse/chemclipse/support/xml/XmlParserFactory.java
```diff
@@ -0,0 +1,70 @@
...
+	private static final String LIMIT_TOTAL_ENTITY_SIZE = "jdk.xml.totalEntitySizeLimit";
+	private static final String NO_LIMIT = "0";
+
+	private XmlParserFactory() {
+
+	}
+
+	/**
+	 * The returned factory is not namespace aware. Call
+	 * {@link DocumentBuilderFactory#setNamespaceAware(boolean)} if the document is
```
Mailaender: AI blurp. That applies to every other setter as well.

akurtakov: Fixed.

Mailaender (approved):
This works with large files in JUnit tests started from the IDE.

akurtakov:
Second commit converts all DocumentBuilderFactory to be create using the new helper to prevent accidentially hitting the problem somewhere.

PR #3113, opened 2026-09-07
Reorganize target preferences

Mailaender:
It has lots of settings, and it was hard to navigate.


Mailaender:
grafik
grafik
grafik
grafik

PR #3114, opened 2026-09-07
Reorganize chromatogram settings

Mailaender:
They are vast and confusing. Some were located in the UI, where they don't do anything. This mostly moves them around into smaller groups. I removed three that were just storing internal values.


Mailaender:
grafik
grafik
grafik
grafik

PR #3116, opened 2026-09-08
Remove --add-modules=ALL-SYSTEM from product file

akurtakov:
Chemclipse doesn't need it.


akurtakov:
PDE adds it by default in https://github.com/eclipse-pde/eclipse.pde/blob/523f495f87650156d3548ecbbc1aba9974c6b5c6/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java#L183 for launch configurations so this PR will not help IDE launches but will remove the "WARNING: Using incubator modules: jdk.incubator.vector" starting the built product directly.

PR #3140, opened 2026-09-14
Update to 2026-09 release

akurtakov:


akurtakov:
It's a generic problem in Dash tool https://github.com/eclipse-dash/dash-licenses/issues/590 . Hopefully some real action will be taken as reducing the batch size is putting workarounds but that's what could be done for outsiders.

PR #3143, opened 2026-09-15
Hide org.eclipse.ui.edit.text.actionSet contributions by default

akurtakov:
Hide org.eclipse.ui.edit.text.actionSet.* as a workaround before Eclipse Platform problem is investigated.


akurtakov:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/4391 should fix the problem on the Platform side.

PR #3158, opened 2026-09-21
Reset the perspectives which are hidden on startup

akurtakov:
Only the perspectives hidden at runtime were marked for reset, hence the placeholders of the shared elements, e.g. the data explorer, were left empty when a perspective was opened for the first time.

Assisted-by: Anthropic Claude Code (claude-opus-5[1m])


akurtakov:
Before this one if one had only welcome perspective saved and clicks on Data perspective tile , Data view was empty.

